PR #30, opened 2024-05-04
Change parquet compression

### LINKED ISSUE #29: GeoParquet output doesn't work with DuckDB
So unfortunately we're using brotli compression, which DuckDB doesn't support. Trying to open LFK-AKTI_EPSG25832.parquet from source cooperative or files I've converted seems to result in `Error: Invalid Error: Unsupported compression codec "BROTLI". Supported options are uncompressed, gzip, snappy or zstd`.

I think zstd is generally recommend, see https://twitter.com/kylebarron2/status/1691483893461839872

cholmes:
Changed default compression from brotli to zstd. This enables the output to work by default with DuckDB, which is a decently important target.

Closes #29

### DIFF
--- a/fiboa_cli/convert_utils.py
+++ b/fiboa_cli/convert_utils.py
@@ -26,7 +26,7 @@ def convert(
         attribution = None,
         store_collection = False,
         license = None,
-        compression = "brotli",
+        compression = "zstd",
         **kwargs):
     """
     Converts a German field boundary datasets to fiboa.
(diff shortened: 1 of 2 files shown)


m-mohr:
I got much better compression with brotli in the German datasets. Maybe we should provide an option? Ideally duckdb implements brotli anyway, basing decisions on a single implementation is not ideal. Is there an issue for brotli in duckdb?

cholmes:
Yeah, I think providing an option would be good. I agree basing decisions on a single implementation is not ideal, but I also think DuckDB is one of the better clients to do things in a 'cloud-native geospatial' way, and that it's good to support as many clients as possible. So I definitely lean towards zstd as the default, with brotli as an option.

PR #31, opened 2024-05-04
Added france eurocrops converter

cholmes:
Adding the Eurocrops version of the french government's 2018 boundaries to the converter.

### DIFF
--- a/fiboa_cli/datasets/fr_ec.py
+++ b/fiboa_cli/datasets/fr_ec.py
@@ -1,0 +1,180 @@
+import re
+import pandas as pd
+
+from ..convert_utils import convert as convert_
+
+# File to read the data from
+# Can read any tabular data format that GeoPandas can read through read_file()
+# Supported protcols: HTTP(S), GCS, S3, or the local file system
+
+# This one doesn't work because the files in the zip are in a folder, not at the root.
+# URI = "https://zenodo.org/records/8229128/files/FR_2018.zip"
+
+# This one doesn't work because the the convert utils doesn't use read_parquet (and most
+# fiona installations won't be built against the parquet library as its not in the wheels).
+# I tried to change the convert utils to use read_parquet, but it didn't work, at least
+# not with this (admittedly large) dataset.
+# URI = "https://data.source.coop/cholmes/eurocrops/geoparquet-projected/FR_2018_EC21.parquet"
+
+# This one should work, but it's a 6 gig file, so it'll take a while to download.
+URI = "https://data.source.coop/cholmes/eurocrops/unprojected/flatgeobuf/FR_2018_EC21.fgb"
+
+# This line I used for calling local files after downloading manually, which worked well.
+#URI = "/Users/cholmes/Downloads/FR_2018/FR_2018_EC21.shp"
+
+# Unique identifier for the collection
+ID = "fr_ec"
+# Title of the collection
+TITLE = "Field boundaries for France from Eurocrops (2018)"
+# Description of the collection. Can be multiline and include CommonMark.
+DESCRIPTION = """
+This dataset contains the field boundaries for all of France in 2018. The data was collected by the French government and harmonized
+by the Eurocrops project and is available on Zenodo.
+
+EuroCrops is a dataset collection combining all publicly available self-declared crop reporting datasets from countries of the European Union.
+The project is funded by the German Space Agency at DLR on behalf of the Federal Ministry for Economic Affairs and Climate Action (BMWK).
+The work is licensed under a [Creative Commons Attribution-ShareAlike 4.0 International License][cc-by-sa].
+
+The project developed a new **Hierarchical Crop and Agriculture Taxonomy (HCAT)** that harmonises all declared crops across the European Union. 
+In the data you'll find this as additional attributes:
+
+| Attribute Name | Explanation                                                 |
+| -------------- | ----------------------------------------------------------- |
+| EC_trans_n     | The original crop name translated into English              |
+| EC_hcat_n      | The machine-readable HCAT name of the crop                  |
+| EC_hcat_c      | The 10-digit HCAT code indicating the hierarchy of the crop |
+"""
+# Bounding box of the data in WGS84 coordinates
+BBOX = [-6.047022416643922, -3.916364769838749, 68.89050422648864, 51.075100624023094] 
+
+# Provider name, can be None if not applicable, must be provided if PROVIDER_URL is provided
+PROVIDER_NAME = "Eurocrops"
+# URL to the homepage of the data or the provider, can be None if not applicable
+PROVIDER_URL = "https://github.com/maja601/EuroCrops/wiki/France"
+# Attribution, can be None if not applicable
+ATTRIBUTION = "Institut National de l'Information Géographique et Forestière"
+
+# License of the data, either
+# 1. a SPDX license identifier (including "dl-de/by-2-0" / "dl-de/zero-2-0"), or
+LICENSE = "CC-BY-2.0"
+# 2. a STAC Link Object with relation type "license"
+# LICENSE = {'title': 'CC-BY-4.0', 'href': 'https://creativecommons.org/licenses/by/4.0/', 'type': 'text/html', 'rel': 'license'}
+
+# Map original column names to fiboa property names
+COLUMNS = {
+    'geometry': 'geometry', #fiboa core field
+    'ID_PARCEL': 'id', #fiboa core field
+    'SURF_PARC': 'area', #fiboa core field
+    'CODE_CULTU': 'code_culture', #fiboa custom field
+    'CODE_GROUP': 'code_group', #fiboa custom field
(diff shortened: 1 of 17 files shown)


fiboa_cli/datasets/fr_ec.py
```diff
@@ -0,0 +1,180 @@
...
+# A list of implemented extension identifiers
+EXTENSIONS = []
+
+# Functions to migrate data in columns to match the fiboa specification.
+# Example: You have a column area_m in square meters and want to convert
+# to hectares as required for the area field in fiboa.
+# Function signature:
+#   func(column: pd.Series) -> pd.Series
+COLUMN_MIGRATIONS = {
+    # replace any value in CULTURE_D2 column with 2018-01-15. This is a complete
```
cholmes, line 85: @m-mohr - do you have a better way to deal with this? Basically all I want to do is set every value in the determination_date column as 2018, since I know that's when the data is from, and the 'collection' fails if I don't have a determination_date (for the collection date range).

Happy to make tickets if there's not something I'm just missing, but it seemed like there wasn't a way to just make a new column and set it with a value. I tried the 'migration' ones but couldn't even get that to work.

Also not sure if we should consider allowing the creation of collections without determination_date? I think the thing to do with this one is maybe to just set a collection level determination date (which is an argument for letting any attribute live at collection or feature level, but is the opposite way from what we've mostly been thinking about).

cholmes: ...And I just saw a @m-mohr slack comment in #fiboa on cloud native geo slack that I suspect solves this:

Try this where the MIGRATION = NONE is in the template:
```
def migrate(gdf):
    gdf["determination_datetime"] = "2020-01-01T00:00:00Z"
    gdf["determination_method"] = "manual"
    return gdf

MIGRATION = migrate
```

And then you also need to add the new columns to the COLUMNS mapping:
```
COLUMNS = {
   # ... whatever you had here before, and add:
    'determination_method': 'determination_method',
    'determination_datetime': 'determination_datetime'
}
```

Glad I didn't actually file an issue ;) 

But I think it'd be good to mention this in the tutorial, and maybe even have an example in the template.

m-mohr: As this came up a couple of times now, I've added a simplification:

```
ADD_COLUMNS = {
  "determination_datetime": "2020-01-01T00:00:00Z",
  "determination_method": "manual"
}
```

fiboa_cli/datasets/fr_ec.py
```diff
@@ -0,0 +1,180 @@
...
+        TITLE,
+        DESCRIPTION,
+        BBOX,
+        provider_name=PROVIDER_NAME,
+        provider_url=PROVIDER_URL,
+        source_coop_url=source_coop_url,
+        extensions=EXTENSIONS,
+        missing_schemas=MISSING_SCHEMAS,
+        column_migrations=COLUMN_MIGRATIONS,
+        column_filters=COLUMN_FILTERS,
```
m-mohr, line 175:
```suggestion
        column_filters=COLUMN_FILTERS,
        column_additions=ADD_COLUMNS,
```

fiboa_cli/datasets/fr_ec.py
```diff
@@ -0,0 +1,180 @@
...
+        },
+        'EC_hcat_c': {
+            'type': 'uint32'
+        }
+    }
+}
+
+
+# Conversion function, usually no changes required
+def convert(output_file, cache_file = None, source_coop_url = None, collection = False):
```
m-mohr, line 134:
```suggestion
def convert(output_file, cache_file = None, source_coop_url = None, collection = False, compression = None):
```

fiboa_cli/datasets/fr_ec.py
```diff
@@ -0,0 +1,180 @@
...
+        provider_url=PROVIDER_URL,
+        source_coop_url=source_coop_url,
+        extensions=EXTENSIONS,
+        missing_schemas=MISSING_SCHEMAS,
+        column_migrations=COLUMN_MIGRATIONS,
+        column_filters=COLUMN_FILTERS,
+        migration=MIGRATION,
+        attribution=ATTRIBUTION,
+        store_collection=collection,
+        license=LICENSE
```
m-mohr, line 179:
```suggestion
        license=LICENSE,
        compression=compression
```

fiboa_cli/datasets/fr_ec.py
```diff
@@ -0,0 +1,180 @@
...
+# Custom function to migrate the GeoDataFrame if the other options are not sufficient
+# This should be the last resort!
+# Function signature:
+#   func(gdf: gpd.GeoDataFrame) -> gpd.GeoDataFrame
+MIGRATION = {
+    # Set the determination datetime to a fixed value of 2018-01-15
+    # Note this didn't work, so I had to use the COLUMN_MIGRATIONS hack above.
+    #'determination_datetime': lambda gdf: gdf.assign(determination_datetime=pd.Timestamp('2018-01-15'))
+
+}
```
m-mohr, line 107:
```suggestion
MIGRATION = None
```

fiboa_cli/datasets/fr_ec.py
```diff
@@ -0,0 +1,180 @@
...
+# LICENSE = {'title': 'CC-BY-4.0', 'href': 'https://creativecommons.org/licenses/by/4.0/', 'type': 'text/html', 'rel': 'license'}
+
+# Map original column names to fiboa property names
+COLUMNS = {
+    'geometry': 'geometry', #fiboa core field
+    'ID_PARCEL': 'id', #fiboa core field
+    'SURF_PARC': 'area', #fiboa core field
+    'CODE_CULTU': 'code_culture', #fiboa custom field
+    'CODE_GROUP': 'code_group', #fiboa custom field
+    'CULTURE_D2': 'determination_datetime', #fiboa core field - this is a hack to get a fixed date in the field.
```
m-mohr, line 70:
```suggestion
```

fiboa_cli/datasets/fr_ec.py
```diff
@@ -0,0 +1,180 @@
...
+by the Eurocrops project and is available on Zenodo.
+
+EuroCrops is a dataset collection combining all publicly available self-declared crop reporting datasets from countries of the European Union.
+The project is funded by the German Space Agency at DLR on behalf of the Federal Ministry for Economic Affairs and Climate Action (BMWK).
+The work is licensed under a [Creative Commons Attribution-ShareAlike 4.0 International License][cc-by-sa].
+
+The project developed a new **Hierarchical Crop and Agriculture Taxonomy (HCAT)** that harmonises all declared crops across the European Union. 
+In the data you'll find this as additional attributes:
+
+| Attribute Name | Explanation                                                 |
```
m-mohr, line 41: Tables are not supported in CommonMark.

fiboa_cli/datasets/fr_ec.py
```diff
@@ -0,0 +1,180 @@
...
+| -------------- | ----------------------------------------------------------- |
+| EC_trans_n     | The original crop name translated into English              |
+| EC_hcat_n      | The machine-readable HCAT name of the crop                  |
+| EC_hcat_c      | The 10-digit HCAT code indicating the hierarchy of the crop |
```
m-mohr, line 45:
```suggestion
- `EC_trans_n`: The original crop name translated into English
- `EC_hcat_n`: The machine-readable HCAT name of the crop
- `EC_hcat_c`: The 10-digit HCAT code indicating the hierarchy of the crop
```

fiboa_cli/datasets/fr_ec.py
```diff
@@ -0,0 +1,180 @@
...
+# fiona installations won't be built against the parquet library as its not in the wheels).
+# I tried to change the convert utils to use read_parquet, but it didn't work, at least
+# not with this (admittedly large) dataset.
+# URI = "https://data.source.coop/cholmes/eurocrops/geoparquet-projected/FR_2018_EC21.parquet"
+
+# This one should work, but it's a 6 gig file, so it'll take a while to download.
+URI = "https://data.source.coop/cholmes/eurocrops/unprojected/flatgeobuf/FR_2018_EC21.fgb"
+
+# This line I used for calling local files after downloading manually, which worked well.
+#URI = "/Users/cholmes/Downloads/FR_2018/FR_2018_EC21.shp"
```
m-mohr, line 23:
```suggestion
```

Use `-c` instead

m-mohr (approved):

cholmes:
> @cholmes Could you please review whether my changes make sense and it works for you as intended? For me it looks good for now.

Yup, looks great! Thanks for all your work on it. Much cleaner now. I'll go ahead and merge.

> PS: I renamed to ec_fr to keep the Eurocrops datasets more closely grouped together, keeping the fr_* scope for official datasets from France.

Cool. How strongly do you feel about fiboa project names and file names following this same convention? The 2 sets of two of two letters seems fine for me for the converter, but I tend to like file names to be a bit more description.

m-mohr:
Not so strongly. We could also name it eurocrops_france or so. Should just be consistent in their 'groups', e.g. all eurocrops ones...

PR #37, opened 2024-05-06
Add ability to read parquet files directly with geopandas

### LINKED ISSUE #32: Support GeoParquet for input for `convert` templates
I was trying to use https://data.source.coop/cholmes/eurocrops/geoparquet-projected/FR_2018_EC21.parquet as input for a france eurocrops converter (see #31), but parquet wasn't recognized. This makes sense as read_file() depends on fiona, and fiona isn't built by default with parquet. Generally the [read_parquet](https://geopandas.org/en/stable/docs/reference/api/geopandas.read_parquet.html) method works better anyways (columnar instead of row based, faster, etc). 

I made a go at checking for .parquet in the filename and using it, but got some weird error. But if that sounds like a reasonable approach I can keep trying it.

cholmes:
Added code to check if a file ends with .parquet and if so then uses the [read_parquet](https://geopandas.org/en/stable/docs/reference/api/geopandas.read_parquet.html) method instead of using fiona (which likely won't have parquet reading).

Closes #32

### DIFF
--- a/fiboa_cli/convert_utils.py
+++ b/fiboa_cli/convert_utils.py
@@ -36,7 +36,7 @@ def convert(
     path = download_file(url, cache_file)
 
     # If file is a parquet file then read with read_parquet
-    if path.endswith(".parquet"):
+    if path.endswith(".parquet") or path.endswith(".geoparquet"):
         gdf = gpd.read_parquet(path, **kwargs)
     else:
         gdf = gpd.read_file(path, **kwargs)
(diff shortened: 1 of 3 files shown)


fiboa_cli/convert_utils.py
```diff
@@ -35,8 +35,11 @@ def convert(
         log("Loading file from: " + url)
     path = download_file(url, cache_file)
 
-    log("Local file is at: " + path)
-    gdf = gpd.read_file(path, **kwargs)
+    # If file is a parquet file then read with read_parquet
+    if path.endswith(".parquet"):
```
m-mohr, line 39:
```suggestion
    if path.endswith(".parquet") or path.endswith(".geoparquet"):
```

m-mohr:
Looks good, thank you.

PR #42, opened 2024-06-05
Add converter for Luxembourg - FieldScapes

snehalchaudhari98:
Adding data converter for the sub-sampled data of the Luxembourg field boundaries. 

Original data : [field bourdaries ](https://data.public.lu/en/datasets/referentiel-des-parcelles-flik/#resources)

- fiboa/data#24


fiboa_cli/datasets/dk.py
```diff
@@ -0,0 +1,163 @@
...
+
+# File to read the data from
+# Can read any tabular data format that GeoPandas can read through read_file()
+# Supported protcols: HTTP(S), GCS, S3, or the local file system
+
+# data original source : https://landbrugsgeodata.fvm.dk/
+URI = "/mnt/c/kerner-lab/fieldscapes/denmark/all_parcels.gpkg" # data seubset for FieldScapes
+
+# Unique identifier for the collection
+ID = "dk"
```
m-mohr, line 19: I think we should prefix the fieldscapes converters with fs_ as done in #43

fiboa_cli/datasets/dk.py
```diff
@@ -0,0 +1,163 @@
...
+# Title of the collection
+TITLE = "Field boundaries for Denmark"
```
m-mohr, line 21:
```suggestion
TITLE = "Field boundaries for Denmark (FieldScapes)"
```

fiboa_cli/datasets/dk.py
```diff
@@ -0,0 +1,163 @@
...
+from ..convert_utils import convert as convert_
+import re
+import pandas as pd
+
+# File to read the data from
+# Can read any tabular data format that GeoPandas can read through read_file()
+# Supported protcols: HTTP(S), GCS, S3, or the local file system
+
+# data original source : https://landbrugsgeodata.fvm.dk/
+URI = "/mnt/c/kerner-lab/fieldscapes/denmark/all_parcels.gpkg" # data seubset for FieldScapes
```
m-mohr, line 16: Same as in #43: I'm not sure how to proceed with this. It's not really useful for the general public if there's no way to get the source data. Any thoughts from the fieldscapes members?

fiboa_cli/datasets/dk.py
```diff
@@ -0,0 +1,163 @@
...
+    'GB_FRADRAG' : 'deduction', # fiboa custom field - translated 
+    'MB_TYPE' : 'block_type', #fiboa custom field
+    'GB_AREAL': 'eligible_area', # fiboa core field i.e. eligible area
+    'geometry':'geometry'# fiboa core field
+}
+
+# Add columns with constant values.
+# The key is the column name, the value is a constant value that's used for all rows.
+ADD_COLUMNS = {
+    "determination_datetime": "2021-01-01T00:00:00Z"
```
m-mohr, line 60: Is this a leftover from the template?

fiboa_cli/datasets/dk.py
```diff
@@ -0,0 +1,163 @@
...
+# Map original column names to fiboa property names
+# You also need to list any column that you may have added in the MIGRATION function (see below).
+COLUMNS = {    
+    'MARKBLOKNR' : 'id', # fiboa core field , TODO MARKBLOKNR-id conversion is needed.     
+    'GEOMETRISK' : 'area',#fiboa custom field i.e total area = eligible area + ineligible area
+    'TARAAREAL': 'ineligible_area', #fiboa custom field
+    'STOETPROC' : 'support_proc',  # fiboa custom field - translated. Supported Scheme. Currently all values are NaN
+    'GB_FRADRAG' : 'deduction', # fiboa custom field - translated 
+    'MB_TYPE' : 'block_type', #fiboa custom field
+    'GB_AREAL': 'eligible_area', # fiboa core field i.e. eligible area
```
m-mohr, line 53: not a core field

fiboa_cli/datasets/dk.py
```diff
@@ -0,0 +1,163 @@
...
+# 2. a STAC Link Object with relation type "license"
+# LICENSE = {"title": "CC-BY-4.0", "href": "https://creativecommons.org/licenses/by/4.0/", "type": "text/html", "rel": "license"}
+
+# Map original column names to fiboa property names
+# You also need to list any column that you may have added in the MIGRATION function (see below).
+COLUMNS = {    
+    'MARKBLOKNR' : 'id', # fiboa core field , TODO MARKBLOKNR-id conversion is needed.     
+    'GEOMETRISK' : 'area',#fiboa custom field i.e total area = eligible area + ineligible area
+    'TARAAREAL': 'ineligible_area', #fiboa custom field
+    'STOETPROC' : 'support_proc',  # fiboa custom field - translated. Supported Scheme. Currently all values are NaN
```
m-mohr, line 50: If all values are NaN, it whould probably be removed for now?

fiboa_cli/datasets/dk.py
```diff
@@ -0,0 +1,163 @@
...
+# 1. a SPDX license identifier (including "dl-de/by-2-0" / "dl-de/zero-2-0"), or
+LICENSE = "CC0-1.0"
+# 2. a STAC Link Object with relation type "license"
+# LICENSE = {"title": "CC-BY-4.0", "href": "https://creativecommons.org/licenses/by/4.0/", "type": "text/html", "rel": "license"}
+
+# Map original column names to fiboa property names
+# You also need to list any column that you may have added in the MIGRATION function (see below).
+COLUMNS = {    
+    'MARKBLOKNR' : 'id', # fiboa core field , TODO MARKBLOKNR-id conversion is needed.     
+    'GEOMETRISK' : 'area',#fiboa custom field i.e total area = eligible area + ineligible area
```
m-mohr, line 48: That's a core field

fiboa_cli/datasets/dk.py
```diff
@@ -0,0 +1,163 @@
...
+# License of the data, either
+# 1. a SPDX license identifier (including "dl-de/by-2-0" / "dl-de/zero-2-0"), or
+LICENSE = "CC0-1.0"
+# 2. a STAC Link Object with relation type "license"
+# LICENSE = {"title": "CC-BY-4.0", "href": "https://creativecommons.org/licenses/by/4.0/", "type": "text/html", "rel": "license"}
+
+# Map original column names to fiboa property names
+# You also need to list any column that you may have added in the MIGRATION function (see below).
+COLUMNS = {    
+    'MARKBLOKNR' : 'id', # fiboa core field , TODO MARKBLOKNR-id conversion is needed.
```
m-mohr, line 47: There's an open todo here.

fiboa_cli/datasets/dk.py
```diff
@@ -0,0 +1,163 @@
...
+            'type': 'float'
+        },
+        'deduction': {
+            'type': 'float'
+        },
+        'support_proc': {
+            'type': 'float'
+        },
+        'block_type': {
+            'type': 'string'
```
m-mohr, line 105: This should probably be an enum?

fiboa_cli/datasets/dk.py
```diff
@@ -0,0 +1,163 @@
...
+# Custom function to migrate the GeoDataFrame if the other options are not sufficient
+# This should be the last resort!
+# Function signature:
+#   func(gdf: gpd.GeoDataFrame) -> gpd.GeoDataFrame
+MIGRATION = None
+
+# Schemas for the fields that are not defined in fiboa
+# Keys must be the values from the COLUMNS dict, not the keys
+MISSING_SCHEMAS = {
+    "properties": {
```
m-mohr, line 91: Should some of the float fields be have a minimum: 0 constraint?

m-mohr (changes requested):
Thanks. I did only check the dk.py file for now, the changes requested for DK in many cases also apply to lu and se though and should be updated accordingly. I'll re-review fully afterwards.

m-mohr:
Please rebase the PR to the latest CLI version, including renaming the URI constant to SOURCES and renaming the parameter cache_file to cache. I tried to push it upstream, but you didn't permit changes to this PR so you need to do these changes yourself. Thanks.

fiboa_cli/datasets/dk.py
cholmes: The core idea here is to get the converter code out asap, but the first goal is to get the resulting files into fiboa on source cooperative. The hope is that @eddiechoi00 will be able to help with adapting the converters to work with the original source data, and then putting the full datasets (instead of the subsets fieldscapes is focused on) in source cooperative.

m-mohr: Yeah, I was thinking the same. We should make it so that the source data can get converted and then there could be an option that people can enable to run on fieldscapes data.

fiboa_cli/datasets/dk.py
snehalchaudhari98: No, so the dataset doesn't have any date attributes for individual parcels but from the metadata, it is possible to identify the year in which the data was collected. So for simplicity, we thought we could choose the template date 01.01.XXXX .

Because in the actual fieldscapes dataset, we have the harvest and plantation season's data from the seasonal crop calendar for the given year.

fiboa_cli/datasets/dk.py
snehalchaudhari98: Yep, will fix this.

fiboa_cli/datasets/dk.py
snehalchaudhari98: Yes, I changed the description in the end thus there was a mixup with the actual area and eligible area.

fiboa_cli/datasets/dk.py
snehalchaudhari98: Yes, good suggestion. We'll add the required constraints to the properties.

PR #51, opened 2024-06-14
Add converter for Belgium / wallonia

ivorbosloper:
Data can not be shared by current license


fiboa_cli/convert_utils.py
```diff
@@ -36,6 +36,9 @@ def convert(
         log("Loading file")
     path = download_file(url, cache_file)
 
+    if not len(bbox) == 4:
+        raise ValueError("Bbox must be of length 4")
+
```
m-mohr, line 41:
```suggestion
```

I've directly added this to the main branch, good idea.

fiboa_cli/datasets/be_wa.py
```diff
@@ -0,0 +1,89 @@
+from ..convert_utils import convert as convert_
+
+URI = "http://use_your_own_copy/"
```
m-mohr, line 3: Let me think about a solution that is a bit better than this approach. Maybe we can let users input the path directly when an empty URL is provided through the template. The -c option will change in #48 anyway.

fiboa_cli/datasets/be_wa.py
```diff
@@ -0,0 +1,89 @@
...
+    'CULT_NOM': 'crop_name',
+    'GROUPE_CULT': 'group_code',
+}
+
+ADD_COLUMNS = {}
+
+# A list of implemented extension identifiers
+EXTENSIONS = []
+
+import pandas as pd
```
m-mohr, line 43: Better move the imports to the top of the file.

tests/data-files/be_wa.zip
m-mohr, line 1: Can we add this here? It seems the license is very restrictive, which would also apply to a subset of the data...

fiboa_cli/convert_utils.py
ivorbosloper: Thanks, I forgot it a few times and this error makes it clear from the start.

tests/data-files/be_wa.zip
ivorbosloper: You're right, I thought a small subsample was acceptable. I've now submitted a test file with completely fake data (only using column names and file name).

m-mohr: Thank you!

PR #57, opened 2024-06-18
Finland: Make valid dataset

ivorbosloper:
Fix data validation issues

### DIFF
--- a/fiboa_cli/datasets/fi.py
+++ b/fiboa_cli/datasets/fi.py
@@ -24,10 +24,16 @@ COLUMNS = {
     "KASVIKOODI_SELITE_FI": "crop_name",
 }
 
+COLUMN_FILTERS = {
+    "PINTA_ALA": lambda col: col > 0.0  # fiboa validator requires area > 0.0
+}
+
 def migrate(gdf):
     # Make year (1st january) from column "VUOSI"
     gdf['determination_datetime'] = pd.to_datetime(gdf['VUOSI'], format='%Y')
+    gdf['geometry'] = gdf["geometry"].make_valid()
     return gdf
+
 
 MISSING_SCHEMAS = {
     "properties": {
@@ -56,6 +62,7 @@ def convert(output_file, cache = None, source_coop_url = None, collection = Fals
         provider_name=PROVIDER_NAME,
         provider_url=PROVIDER_URL,
         source_coop_url=source_coop_url,
+        column_filters=COLUMN_FILTERS,
         missing_schemas=MISSING_SCHEMAS,
         migration=migrate,
         attribution=ATTRIBUTION,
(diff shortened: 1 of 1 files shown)


fiboa_cli/datasets/fi.py
```diff
@@ -24,11 +24,17 @@
     "KASVIKOODI_SELITE_FI": "crop_name",
 }
 
+COLUMN_FILTERS = {
+    "PINTA_ALA": lambda col: col > 0.0  # fiboa validator requires area > 0.0
```
m-mohr, line 28: Would it make more sense to set invalid values to None/Null instead of filtering the data out? Not sure whether the fields are actually invalid or just have invalid areas computed...

ivorbosloper: Yes, I've created an alternative...

m-mohr:
Thanks. The question is whether we want to actually "enrich" the data. If we encounter something invalid maybe we should just set it to null? Or should we try to be smart and compute new values?

ivorbosloper:
> Thanks. The question is whether we want to actually "enrich" the data. If we encounter something invalid maybe we should just set it to null? Or should we try to be smart and compute new values?

This dataset contains an _area_ attribute which has been rounded to 2 decimals, resulting in quite some rows with an area of 0.00 . The `fiboa validate` does not allow these values for the derived parquet-file. Would _null_ be allowed by the validator? And is _null_ is more correct than a calculated area? As a dataset user, I would prefer rows with a calculated_area, but I can live with _null_ as well...

m-mohr:
> This dataset contains an area attribute which has been rounded to 2 decimals, 

Ah! This bit of information was missing for me. I was under the assumption it might just be invalid values. With this context, I'm happy with the implementation.

fiboa_cli/datasets/fi.py
```diff
@@ -19,16 +20,22 @@
...
     "PERUSLOHKOTUNNUS": "id",
     "LOHKONUMERO": "block_id",
-    "PINTA_ALA": "area",
+    "area": "area",
     "KASVIKOODI": "crop_code",
     "KASVIKOODI_SELITE_FI": "crop_name",
 }
 
+COLUMN_FILTERS = {}
+
```
m-mohr, line 29:
```suggestion
```

fiboa_cli/datasets/fi.py
```diff
@@ -56,6 +63,7 @@ def convert(output_file, cache = None, source_coop_url = None, collection = Fals
         provider_name=PROVIDER_NAME,
         provider_url=PROVIDER_URL,
         source_coop_url=source_coop_url,
+        column_filters=COLUMN_FILTERS,
         missing_schemas=MISSING_SCHEMAS,
```
m-mohr, line 67:
```suggestion
        missing_schemas=MISSING_SCHEMAS,
```

PR #58, opened 2024-06-18
Automate publication process

ivorbosloper:
Example: `fiboa publish fi -e finland -c ~/Downloads fi_data`

Automates https://github.com/fiboa/data/blob/main/HOWTO.md#each-time-you-update-the-dataset step 9-19

### DIFF
--- a/fiboa_cli/__init__.py
+++ b/fiboa_cli/__init__.py
@@ -15,6 +15,7 @@ from .validate import validate as validate_
 from .validate_schema import validate_schema as validate_schema_
 from .version import __version__, fiboa_version as fiboa_version_
 from .util import log, check_ext_schema_for_cli, valid_file_for_cli, valid_file_for_cli_with_ext, valid_files_folders_for_cli, valid_folder_for_cli
+from .publish import publish as publish_
 
 @click.group()
 @click.version_option(version=__version__)
@@ -427,6 +428,36 @@ def rename_extension(folder, title, slug, org = "fiboa", prefix = None):
         sys.exit(1)
 
 
+## Publish
+@click.command()
+@click.argument('dataset', nargs=1, type=click.Choice(list_all_converter_ids()))
+@click.argument('directory', nargs=1, type=click.Path(exists=False))
+@click.option(
+    '--cache', '-c',
+    type=click.Path(exists=False),
+    help='By default the CLI downloads the source data on every execution. Specify a local file path to avoid downloading the file again. If the file exists, reads from the path, otherwise stores the file there.',
+    default=None
+)
+@click.option(
+    '--source-coop-extension', '-e',
+    type=click.STRING,
+    help='(Future) source_coop extension, will be used as https://beta.source.coop/fiboa/xx-yy/',
+    default=None
+)
+def publish(dataset, directory, cache, source_coop_extension):
+    """
+    Publish a fiboa collection on
+    """
+    log(f"Trying to publish on source coop CLI {__version__}\n", "success")
+    try:
+        directory = os.path.abspath(directory)
+        publish_(dataset, directory, cache, source_coop_extension)
+        log(f"Dataset published from {directory}", "success")
+    except Exception as e:
+        log(e, "error")
+        sys.exit(1)
+
+
 cli.add_command(describe)
 cli.add_command(validate)
 cli.add_command(validate_schema)
@@ -436,6 +467,7 @@ cli.add_command(jsonschema)
 cli.add_command(convert)
 cli.add_command(converters)
 cli.add_command(rename_extension)
+cli.add_command(publish)
 
 if __name__ == '__main__':
     cli()
--- a/fiboa_cli/publish.py
+++ b/fiboa_cli/publish.py
@@ -1,0 +1,97 @@
+import os
+import sys
+import json
+import requests
+
+from .util import log
+from fiboa_cli import list_all_converter_ids
+from fiboa_cli.convert import convert
+from fiboa_cli.validate import validate
+
+STAC_EXTENSION = "https://stac-extensions.github.io/web-map-links/v1.2.0/schema.json"
+
+
+def exc(cmd):
+    assert os.system(cmd) == 0
+
+
+def publish(dataset, directory, cache, source_coop_extension):
+    """
+    Implement https://github.com/fiboa/data/blob/main/HOWTO.md#each-time-you-update-the-dataset
+    """
+    assert dataset in list_all_converter_ids()
+    os.makedirs(directory, exist_ok=True)
+
+    for key in ("README.md", "LICENSE.txt"):
+        path = os.path.join(directory, key)
+        if not os.path.exists(path):
+            raise FileNotFoundError(f"Please create {path}")
+
+    parent = os.path.dirname(directory)
+    os.chdir(parent)
+
+    if not source_coop_extension:
+        source_coop_extension = dataset.replace("_", "-")
+
+    file_name = source_coop_extension.replace("-", "_")  # not sure if we want this
+
+    parquet_file = os.path.join(directory, f"{file_name}.parquet")
+    source_coop_url = f"https://beta.source.coop/fiboa/{source_coop_extension}/"
+
+    assert requests.get(f"https://api.source.coop/repositories/fiboa/{source_coop_extension}").status_code == 200, \
+        f"Missing repo at {source_coop_url}"
+
+    collection_file = os.path.join(directory, "collection.json")
+
+    stac_directory = os.path.join(directory, "stac")
+    done_convert = os.path.exists(parquet_file) and os.path.exists(os.path.join(stac_directory, "collection.json"))
+
+    if not done_convert:
+        # fiboa convert xx_yy -o data/xx-yy.parquet -h https://beta.source.coop/fiboa/xx-yy/ --collection
+        log(f"Converting file for {dataset} to {parquet_file}\n", "success")
+        convert(dataset, parquet_file, cache, source_coop_url, collection=True)
+        log(f"Done\n", "success")
+    else:
+        log(f"Using existing file {parquet_file} for {dataset}\n", "success")
+
+    # fiboa validate data/xx-yy.parquet --data
+    log(f"Validating {parquet_file}", "info")
+    result = validate(parquet_file, config={"data": True})
+    if result:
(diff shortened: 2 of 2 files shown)


cholmes:
This is so awesome!!! I've been dreaming about something like this. If I have time I'll test it out later today.

This won't actually make the readme file for you though, right? I was thinking that could be a nice thing to do too, likely as a separate command - generate the nice table view of the values from the converter or fiboa schema in some way.

fiboa_cli/publish.py
```diff
@@ -0,0 +1,97 @@
...
+        exc(f"ogr2ogr -t_srs EPSG:4326 geo.json {parquet_file}")
+
+        log(f"Running tippecanoe", "info")
+        exc(f"tippecanoe -zg --projection=EPSG:4326 -o {pm_file} -l {dataset} geo.json --drop-densest-as-needed")
+
+        os.remove("geo.json")
+
+    log(f"Uploading to aws", "info")
+    log(f"Get your credentials at {source_coop_url}download/ (press ACCESS DATA and GENERATE CREDENTIALS", "info")
+    exc(f"aws s3 sync {directory} s3://us-west-2.opendata.source.coop/fiboa/{source_coop_extension}/")
```
m-mohr, line 97: I'm just wondering, but shouldn't the CLI stop after the log and let users input the credentials?

Also, we are backing away from using the fiboa org and aim at users hosting under their own accounts, so this may need some slight changes.

fiboa_cli/publish.py
```diff
@@ -0,0 +1,97 @@
...
+            json.dump(data, f)
+        os.remove(collection_file)
+
+    pm_file = os.path.join(directory, f"{file_name}.pmtiles")
+    if not os.path.exists(pm_file):
+        log(f"ogr2ogr geo.json", "info")
+        exc(f"ogr2ogr -t_srs EPSG:4326 geo.json {parquet_file}")
+
+        log(f"Running tippecanoe", "info")
+        exc(f"tippecanoe -zg --projection=EPSG:4326 -o {pm_file} -l {dataset} geo.json --drop-densest-as-needed")
```
m-mohr, line 91: This has some heavy dependencies that would need to be installed somehow. Running into GDAL installation issues everytime, I feel like this might be hard to solve properly...

cholmes: Maybe have these ones be a little more 'opt-in'? Like we just check if the libraries are installed, and if not it just says 'please install GDAL and tippecanoe', and skips the steps.

ivorbosloper: Thanks for the valid points in the review. I realized this pull-request was a work-in-progress that "works on my machine", but wanted to check if it was an interesting feature for the project. 

To get a data-set published/updated, you need to follow [the HOWTO](https://github.com/fiboa/data/blob/main/HOWTO.md#each-time-you-update-the-dataset) and you will need gdal & tippecanoe on your machine. I relied on `os.system()` calls as opposed to python-libraries to avoid introducing more dependencies. Whenever I see a list of detailed instructions like in the HOWTO, I'm always afraid I (or one of my project-members) will mess up / forget something. A script like this -with enough guidance and sanity checks- will keep us all safe. Let's see if I can get some more sanity checks in there...

fiboa_cli/publish.py
ivorbosloper: Only if it fails. After generating credentials, source coop shows a pre-filled _Usage Examples_ section with export-statements. Copy-paste in your terminal and re-run the publish command.

```
export AWS_DEFAULT_REGION=us-west-2
export AWS_ACCESS_KEY_ID=<filled out id>
export AWS_SECRET_ACCESS_KEY=<filled out key>
export AWS_SESSION_TOKEN=<filled out token>

fiboa publish fi -e finland -c ~/Downloads fi_data
```

ivorbosloper: I added some documentation / user instructions...

fiboa_cli/publish.py
m-mohr: I think it's a good idea. I think this might need to be a bit more interactive, giving instructions to users and waiting for confirmation, ... - more like a step-by-step wizard ;-)

ivorbosloper: Ok. @m-mohr I'm not sure what part you're considering making interactive. We could pre-generate the README and LICENSE by scraping the specific page from the [data-survey](https://raw.githubusercontent.com/fiboa/data-survey/main/data/FR.md). Or we could request a copy/paste-input of secrets from source coop. 

Currently the script stops with some instructions (write README, install tippecanoe, copy the `export AWS_` commands in the terminal.

cholmes: I think he meant just meant interactive with the requirements for the 'additional things' - like prompting someone 'would you like to generate pmtiles y/n?', and then have it check if tippecanoe is installed, and if not tell the user 'tippecanoe must be installed to generate pmtiles, do XYZ to install it'.

setup.py
```diff
@@ -41,6 +41,9 @@ def get_description():
         # ]
         "ie": [
             "zipfile-deflate64"
+        ],
+        "si": [
+            "rarfile>=4.0"
```
m-mohr, line 46: This and the si.py are probably unintentionally in here?

fiboa_cli/publish.py
```diff
@@ -0,0 +1,200 @@
...
+It has been converted to a fiboa GeoParquet file from data obtained from {props['Data Provider (Legal Entity)']}.
+
+- **Source Data Provider:** [{props['Data Provider (Legal Entity)']}]({props['Homepage']})
+- **Converted by:** {props['Submitter (Affiliation)']}
+- **License:** {props['License']}
+- **Projection:** {props['Projection']}
+
+---
+
+- **[Download the data as fiboa GeoParquet]({source_coop_url}/{file_name}.parquet)
```
m-mohr, line 80: Either

```suggestion
- [Download the data as fiboa GeoParquet]({source_coop_url}/{file_name}.parquet)
```

```suggestion
- **[Download the data as fiboa GeoParquet]({source_coop_url}/{file_name}.parquet)**
```

fiboa_cli/publish.py
```diff
@@ -0,0 +1,200 @@
...
+        exc(f"tippecanoe -zg --projection=EPSG:4326 -o {pm_file} -l {dataset} geo.json --drop-densest-as-needed")
+
+        os.remove("geo.json")
+
+    log(f"Uploading to aws", "info")
+    if not os.environ.get("AWS_SECRET_ACCESS_KEY"):
+        log(f"Get your credentials at {source_coop_url}manage/", "info")
+        log("  Then press 'ACCESS DATA',\n  and click 'Create API Key',", "info")
+        log("  Run export AWS_DEFAULT_REGION=us-west-2 AWS_ACCESS_KEY_ID=<> AWS_SECRET_ACCESS_KEY=<>\n"
+            "  where you copy-past the access key and secret", "info")
```
m-mohr, line 194: That's outdated

fiboa_cli/publish.py
```diff
@@ -0,0 +1,200 @@
...
+- Data downloaded on {date.today()} from {next(iter(converter.SOURCES))} .
+- Converted to GeoParquet using [fiboa-cli](https://github.com/fiboa/cli), version {version.__version__}
+"""
+
+
+def publish(dataset, directory, cache, source_coop_extension):
+    """
+    Implement https://github.com/fiboa/data/blob/main/HOWTO.md#each-time-you-update-the-dataset
+
+    You need GDAL 3.8 or later (for ogr2ogr), tippecanoe, and AWS CLI
```
m-mohr, line 100: You also need libgdal-arrow-parquet

setup.py
ivorbosloper: Pfiew. I messed up somewhere with rebasing 🙈

fiboa_cli/publish.py
```diff
@@ -190,7 +190,7 @@ def publish(dataset, directory, cache, source_coop_extension):
     if not os.environ.get("AWS_SECRET_ACCESS_KEY"):
         log(f"Get your credentials at {source_coop_url}manage/", "info")
         log("  Then press 'ACCESS DATA',\n  and click 'Create API Key',", "info")
-        log("  Run export AWS_DEFAULT_REGION=us-west-2 AWS_ACCESS_KEY_ID=<> AWS_SECRET_ACCESS_KEY=<>\n"
+        log("  Run export AWS_DEFAULT_REGION=https://data.source.coop AWS_ACCESS_KEY_ID=<> AWS_SECRET_ACCESS_KEY=<>\n"
```
m-mohr, line 193: Shouldn't that be the endpoint URL instead of the region?

ivorbosloper: Yes, thanks for checking

fiboa_cli/__init__.py
```diff
@@ -503,6 +504,9 @@ def rename_extension(folder, title, slug, org = "fiboa", prefix = None):
     except Exception as e:
         log(e, "error")
         sys.exit(1)
+    except Exception as e:
+        log(e, "error")
+        sys.exit(1)
```
m-mohr, line 509: Looks like a merge issue...

cholmes (commented):
It would be another dependency, but I've found the best way to convert GeoParquet to PMTiles is to use GPQ. See https://gist.github.com/cholmes/d086d7d5ca4ac0148aa2f1259fd813d8

Then you don't have to create geojson or flatgeobuf. It is a small dependency, and runs well cross platform, though I think the user would have to install it themselves....

m-mohr:
fyi: pipes don't work on Windows so this makes it only work on *nix systems, which we should try to avoid ;-)

cholmes:
> Is gpq somehow better than ogr2ogr?

It's a lot smaller / cleaner than ogr2ogr - but if we're already requiring gdal/ogr then it doesn't offer much advantage.

> fyi: pipes don't work on Windows, so I'd try to avoid it ;-)

Really? Ugh. Yeah, guess we should avoid the pipes then...

> We have a GeoJSON converter in fiboa CLI: fiboa create-geojson example.parquet -o dest-folder

Well, I think the issue Ivor was trying to solve is to not create huge geojson files to drive the pmtiles creation. If we could use pipes and get that command to pipe as input to tippecanoe then it'd be a great solution. But if we're avoiding pipes then not so much, and going to flatgeobuf will save space.

PR #59, opened 2024-06-20
initial attempt at planet data converter

cholmes:
This one needs a bit of help, but I think is pretty close.

I got it mostly working, though couldn't figure out why it was using 0.4.0 of the CLI, until I pushed the branch and realized I'd accidentally branched off of the Austria and Brazil converter branch that I was investigating. But I had managed to get it working locally.

Then I switched to 0.5.0, pleased to remove my dummy 'bbox' that I stuck in, but it's now stuck. I get:

```
% fiboa convert planet_afb -o ~/repos/data-fiboa/output/planet.parquet
fiboa CLI 0.5.0 - Convert 'planet_afb'

Getting file(s) if not cached yet
'dict' object has no attribute 'delete'
```

I couldn't find where the 'delete' was coming from, or where it should be. Any ideas? 

This PR also raises some issues, I mostly highlighted as comments in the code, the main one being that this is not about converted a fixed, online dataset, but is meant to be a tool to converter any Planet field boundary output. So it'd ideally have like a `-i` flag to point at a file instead of a default online location, or even require an input argument for this conversion. So curious how to approach this. It will be similar for DigiFarm and OneSoil, two other converters I'd like to make, as those are the three main field boundary providers that I know of.

### DIFF
--- a/fiboa_cli/datasets/planet_afb.py
+++ b/fiboa_cli/datasets/planet_afb.py
@@ -7,13 +7,19 @@ import re
 SOURCES = None
 
 DATA_ACCESS = """
-Get the data from Planet through ...
+Data must be obtained from the Planet subscriptions API, see 
+https://developers.planet.com/docs/planetary-variables/field-boundaries/ for additional information.
+The output should look something like FIELD_BOUNDARIES_v1.0.0_S2_P1M-20230101T000000Z_fb.gpkg
 """
 
 ID = "planet_afb"
 SHORT_NAME = "Planet Field Boundaries"
 TITLE = "Field boundaries created by Planet's Automated Field Boundary detection algorithm"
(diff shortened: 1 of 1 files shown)


m-mohr:
The delete issue has been fixed, I think. It only appears with Python 3.12 or later, which I don't have yet.

With the changes relating to extracting ZIP files, handling multiple input files, and caching, it's not overly easy anymore to pass in the source files easily. The -c option which we've used before now accepts a folder with files included and then the converter needs to specify the filenames in the folders. Also, a -i option could only handle the simple usecase of a non-extractable files.
Have a look at be_wa, which has also external data that you need to pass in from an external source. There's a DATA_ACCESS flag, which you can fill with instructions and then you can use -c and SOURCES in combination. Would that work for you?

cholmes:
Ok, I checked the improvements, they all look great. I updated the docs on this a bit, and also added determination method. So I think it's ready to merge.

thanks!

m-mohr (approved):

PR #60, opened 2024-06-21
Converter for Portugal government data

ivorbosloper:


fiboa_cli/datasets/pt.py
```diff
@@ -0,0 +1,84 @@
...
+    # filter for layers starting with Culturas_
+    gdfs = []
+    layers = fiona.listlayers(path)
+    for layer in layers:
+        if not layer.startswith("Culturas_"):
+            continue
+        data = gpd.read_file(path, layer=layer)
+        gdfs.append(data)
+
+    return pd.concat(gdfs)
```
m-mohr, line 61: All data should already be loaded, so I think this needs to happen before to avoid wasting resources.
The file migration is more for merging different types of data, let me find a solution that is more integrated...

fiboa_cli/datasets/pt.py
```diff
@@ -0,0 +1,84 @@
...
+# Inspire license. Not 100% clear at source
+LICENSE = {"title": "No conditions apply", "href": "https://inspire.ec.europa.eu/metadata-codelist/ConditionsApplyingToAccessAndUse/noConditionsApply", "type": "text/html", "rel": "license"}
+
+COLUMNS = {
+    "geometry": "geometry",
+    "OSA_ID": "id",
+    "CUL_ID": "block_id",
+    "CUL_CODIGO": "crop_code",
+    "CT_português": "crop_name",
+    "Shape_Area": "area"
```
m-mohr, line 25: Can we make use of the Shape_Length as well? For perimeter...

fiboa_cli/datasets/pt.py
m-mohr: Updated the CLI code to be able to filter by layers. Updated your template accordingly. Please check whether you are happy with it.

m-mohr (approved):
Generally, looks good to me. Please check the two comments and then happy to merge once you've confirmed.

fiboa_cli/datasets/pt.py
ivorbosloper: Ok, I think the layer-filter is a useful feature. I have some doubts about a bug in the current implementation (handling of .parquet files): https://github.com/fiboa/cli/commit/49b54df29c7bbcf93e67d5668017c35d8c66bea9#r143467019

Currently, `_convert` is a complex function with many arguments. In a similar setup, I've used some kind of `Converter` super-class, overridable by all converters. This makes it easier to implement some custom unforeseen overrides. But for now this is workable.

fiboa_cli/datasets/pt.py
ivorbosloper: Thanks. I was unaware of the usefulness, added.

fiboa_cli/datasets/pt.py
m-mohr: Good catch, thank you!
Fixed it in https://github.com/fiboa/cli/commit/f1140a60931baceaf1deee44322d235eb49bf0bd
Unfortunately, we don't have a GeoParquet source in converters anymore, so there's no test that catches this...

I agree that the convert function has grown significantly and should be refactored. Before I do this though, we should add more tests so that we can ensure a refactored version works properly. I generally started with testing various other commands today and I'll somewhat require tests for converters soon: #72 
Having that said, we should probably also test the converters a little more. Just relying on the converter in general, doesn't necessarily identifies "bugs" in the data.

PR #64, opened 2024-06-22
Update ec_si.py template

DanielNwaeze:
Removed determination datetime column and added EC_nuts3 in columns list with description

### DIFF
--- a/fiboa_cli/datasets/ec_si.py
+++ b/fiboa_cli/datasets/ec_si.py
@@ -1,12 +1,23 @@
+import os
+
+from tempfile import TemporaryDirectory
+from zipfile import ZipFile
+
+from ..util import download_file, log
 from ..convert_utils import convert as convert_
 
-SOURCES = {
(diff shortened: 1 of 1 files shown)


m-mohr (changes requested):
There are many unrelated changes in the PR that override recent commits. Please clean this up.

DanielNwaeze:
@m-mohr There seems to be a lot of changes you've made in the Fiboa repo and now, running fiboa convert is broken. Did we really need to make these modifications?

DanielNwaeze:
No worries. It was a dependency issue as I can see the cli version was updated

m-mohr:
Yes, we needed the changes, you just would have needed to work on the latest version. Why did you not `git pull` before you applied further updates? @DanielNwaeze

DanielNwaeze:
I had no idea the version got updated on Git. it was the same as at Thursday when I pulled it. Logged on yesterday with the assumption it was still the same. Oversight on my part

PR #68, opened 2024-06-23
Fieldscapes converter for Belgium and France

snehalchaudhari98:
Adding data converter for the sub-sampled data of the following countries.

References :

- fiboa/data#20
    - [Belgium Field Boundaries](https://landbouwcijfers.vlaanderen.be/open-geodata-landbouwgebruikspercelen)
- fiboa/data#14
    -  [France Field Boundaries](https://geoservices.ign.fr/rpg#telechargementrpg2021)


cholmes (commented):
Note that I put up a converter for eurocrops france as well, at https://github.com/fiboa/cli/blob/main/fiboa_cli/datasets/ec_fr.py. I think this one may be a bit better, but we should be sure we just have one in the repo.

snehalchaudhari98:
- fiboa/specification#28 this issue is present in the Franceconverter.  As per earlier discussions, for the fieldscapes project, the original geometries should not be changed.

PR #71, opened 2024-06-23
Update ec_si.py

DanielNwaeze:
Modified based on latest cli changes for commit

### DIFF
--- a/fiboa_cli/datasets/ec_si.py
+++ b/fiboa_cli/datasets/ec_si.py
@@ -46,7 +46,7 @@ COLUMNS = {
     'EC_trans_n': 'EC_trans_n', #fiboa custom field
     'EC_hcat_n': 'EC_hcat_n', #fiboa custom field
     'EC_hcat_c': 'EC_hcat_c', #fiboa custom field
-    'EC_NUTS3': 'EC_nuts3', #fiboa custom field
+    'EC_NUTS3': 'EC_NUTS3', #fiboa custom field
 }
 
 ADD_COLUMNS = {}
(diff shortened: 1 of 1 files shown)


fiboa_cli/datasets/ec_si.py
```diff
@@ -40,15 +45,14 @@
     'COLOR': 'color', #fiboa custom field
     'EC_trans_n': 'EC_trans_n', #fiboa custom field
     'EC_hcat_n': 'EC_hcat_n', #fiboa custom field
-    'EC_hcat_c': 'EC_hcat_c' #fiboa custom field
+    'EC_hcat_c': 'EC_hcat_c', #fiboa custom field
+    'EC_NUTS3': 'EC_nuts3', #fiboa custom field
```
m-mohr, line 49: If other EC datasets also use this field, maybe keep it uppercase?

DanielNwaeze (commented):
Modified the change @m-mohr

m-mohr:
While you are at it, would you mind adding a test similar to the ones we have?

Will merge nevertheless now, changes look good.

DanielNwaeze:
I don't see steps to creating tests on the repo. Maybe we add something clear in the Readme?

m-mohr:
That's something relatively new, we should indeed update the Readme.

DanielNwaeze:
what do we include for the data files for eurocrops datasets ec_fr and ec_si? no gpkgs are contained in the data.
do we convert the .shp file to .gpkg file? and wouldn't that consume a large amount of data?

m-mohr:
We can create a subset of the data using ogr. See the test file for the command.

PR #75, opened 2024-06-27
Added tests for Slovenia Eurocrops' dataset

DanielNwaeze:
Added `ec_si` test gpkg file and modified `test_convert.py` to include `ec_si`


DanielNwaeze:
@m-mohr error says:

>  area: Value 140359.53125 is greater than the maximum allowed value of 100000.
Extend the area max value?

m-mohr:
I'd assume the value is invalid or in the wrong unit, could you check that?

DanielNwaeze:
I can confirm that the unit of area in Slovenia is in square metre. I assume the default unit for fiboa tests is hectare. this means i'll modify the template using `column_migrations` to multiply the area by 0.0001 to make it in hectare

m-mohr:
I've merged #76, thanks. Can you update this branch so that we can check the CI for this PR?

DanielNwaeze:
Checks passed now!

PR #78, opened 2024-06-28
updates to planet converter to use the planet fiboa extension

cholmes:
Updated the planet_afb converter to try to use the extension. It seems to mostly work, but I get 

```
planet:ca_ratio: No schema defined, converting float64 to nullable double
planet:micd: No schema defined, converting float64 to nullable double
planet:qa: No schema defined, converting int64 to nullable int64
```

when I run the conversion - `fiboa convert planet_afb -i ~/repos/data-fiboa/source/FIELD_BOUNDARIES_v1.0.0_S2_P1M-20230101T000000Z_fb.gpkg -o ~/Downloads/updated-planet-fiboa.parquet`

It seems to use the extension at least in some way, but then doesn't seem to pick up what types to use? I probably missed something I'm supposed to tweak - help appreciated.

### DIFF
--- a/fiboa_cli/datasets/planet_afb.py
+++ b/fiboa_cli/datasets/planet_afb.py
@@ -55,8 +55,6 @@ ADD_COLUMNS = {
   "determination_method": "auto-imagery"
 }
 
-EXTENSIONS = []
-
 def FILE_MIGRATION(gdf, path, uri, layer = None):
     # The file name contains the date, so we can use that to add a
     # date column to the dataset.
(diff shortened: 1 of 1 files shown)


m-mohr:
@cholmes You are overriding EXTENSIONS in line 58 with an empty array :-)

cholmes:
Thanks! That definitely fixed it. Looked way to quickly for that section...

m-mohr (approved):

PR #80, opened 2024-07-14
eurocrops latvia

eddiechoi00:


m-mohr:
43 changed files? Somethings wrong here... 🤔

eddiechoi00:
@m-mohr Yea, I'm really not sure what happened here – I cloned the repository, checked out a new branch, added the necessary files, committed then pushed it, but it seems to think I've changed every single file like it was an empty repository before...

m-mohr:
They are all in a separate folder though, build. I think you could simply remove it?

eddiechoi00:
My apologies, I didn't notice they were in a separate folder. I think they were created when I was committing.

eddiechoi00:
@m-mohr I've deleted the files.

m-mohr:
Great. Looks good. What's up with PRODUCT_DE?

m-mohr:
Has it exactly the same values? If yes, you can likely leave it out, indeed.

eddiechoi00:
They seem to be exactly same except for 1 value of 432188 values.
The said difference exists where one value of `PRODUCT_DE` for all values of `ec:translated_name` equals to **'Meadow fescue for seed production'** is listed as **'P?avas skarene, s?klas ieguvei'** instead of **'P?avas auzene, s?klas ieguvei'**. Considering the encoding issues of `PRODUCT_DE`, I think it would be better to leave it out despite this issue. What do you think?

PR #81, opened 2024-07-19
first attempt at digifarm converter

cholmes:
First attempt at Digifarm converter. This requires users to make an API call and download the resulting GeoJSON. I think this should be good enough for a first contribution.

I was hoping that I could directly call the DigiFarm API, like https://api.digifarm.io/v1/delineated-fields?token=a0731a8c-5259-4c68-af3a-7ad4f6d53faa&bbox=11.13,60.72,11.21,60.76 as the 'input', but it looks like our framework requires a filename for the API. @m-mohr - not sure what you think about relaxing that in some way, or other ideas of how to handle it. I think it'd save a step for people - right now they have to save the file locally and then read it. 

I put some notes in the file itself, but I think it'd be even nicer if the CLI could make a request like `fiboa convert digifarm --bbox 11.13,60.72,11.21,60.76 --token 3a-7ad4f6d53faa`. But that starts to overload the converter a lot, so I also suggested another tool like `fiboa api-request digifarm --bbox 11.13,60.72,11.21,60.76 --token 3a-7ad4f6d53faa | fiboa convert digifarm -i -`. I think OneSoil API would benefit from a similar thing. Planet doesn't have a full API yet, so the file mechanism works fine. (We can break this out as its own issue, just wanted to start the context here).

### DIFF
--- a/fiboa_cli/datasets/digifarm.py
+++ b/fiboa_cli/datasets/digifarm.py
@@ -7,42 +7,21 @@
 
 from ..convert_utils import convert as convert_
 
-# File(s) to read the data from. This should be a GeoJSON file saved from the Digifarm API. 
-#
-# I'm including a 'source' commented out for testing, but setting source to 'None' so
-# that the user knows they need to get the data from the API.
-# SOURCES = "https://gist.githubusercontent.com/cholmes/de82be680b6f9a14ea9e61cf5cbd9c91/raw/2b30dd30a6ed409c23f5181272e20ef3acd4808e/digifarm-norway-sample.json"
 SOURCES = None
-
 DATA_ACCESS = """
-Data must be obtained from the Digifarm API, see https://api-docs.digifarm.io/. Just save any
-API output to a GeoJSON file (with .json as the file extension) and use that as the input file.
-If you want some DigiFarm data, anyone can use:
-https://api.digifarm.io/v1/delineated-fields?token=a0731a8c-5259-4c68-af3a-7ad4f6d53faa&bbox=11.13,60.72,11.21,60.76
-Just save the result as something like 'digifarm-sample.json' and then use -i with the file name to convert it.
+Data must be obtained from the Digifarm API, see https://api-docs.digifarm.io/.
+Use the `-i` CLI parameter to provide the data source.
+Provide a URL to an API request (e.g. `https://api.digifarm.io/v1/delineated-fields?token=...&bbox=11.13,60.72,11.21,60.76`)
+or download the API response to a local GeoJSON file (use `.json` as file extension).
 """
 
-# A filter function for the layer in the file(s) to read.
-# Set to None if the file contains only one layer or all layers should be read.
-# Function signature:
-#   func(layer: str, path: str) -> bool
-LAYER_FILTER = None
-
-# Unique identifier for the collection
 ID = "digifarm"
-# Geonames for the data (e.g. Country, Region, Source, Year)
-SHORT_NAME = "DigiFarm Field Boundaries"
-# Title of the collection
+SHORT_NAME = "DigiFarm"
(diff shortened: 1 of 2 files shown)


cholmes:
Digifarm also sometimes does geopackage output, it seems like the converters should work fine with that - both just get read with geopandas, right?

@m-mohr - if there's additional potential attributes, like 'flatness' in the geopackage should I put those as optional columns? I'm hoping that if the geojson or geopackage doesn't have those then it'd just ignore them if they're not required, but would include them if they are listed as 'optional'.

cholmes:
@m-mohr - tried this out with some bulk shapefile data from digifarm. See https://github.com/fiboa/cli/commit/ae7b088d79c99b9e249e5f162f9723f43b8095d4

It's nice that it mostly works - these converters can take in more than just a single data format as input. But I got issues with 'AREA'. Not sure if that's because it's shapefile, or digifarm converted the data that way. But it wondering if we should make the converters case insensitive? My first attempt it didn't convert 'area' because it read it in as 'AREA'.

fiboa_cli/datasets/digifarm.py
```diff
@@ -0,0 +1,146 @@
...
+# 2. a STAC Link Object with relation type "license"
+LICENSE = {"title": "Proprietary", "href": "https://digifarm.io/legal/tc", "type": "text/html", "rel": "license"}
+
+# Map original column names to fiboa property names
+# You also need to list any column that you may have added in the MIGRATION function (see below).
+COLUMNS = {
+    "area_ha": "area",
+    "id": "id",
+    "geometry": "geometry",
+    "area": "area_sq_m",
```
m-mohr, line 71: We use hectares, not m²

m-mohr:
I fixed the issue with `-i`, you can now use a URL as input.

For `fiboa api-request` we should have a separate issue, indeed.

> if there's additional potential attributes, like 'flatness' in the geopackage should I put those as optional columns? 

There's no real strategy available for this, but you could add the properties and whenever they are not available the converter throws a warning.

> . But I got issues with 'AREA'. Not sure if that's because it's shapefile, or digifarm converted the data that way. But it wondering if we should make the converters case insensitive? My first attempt it didn't convert 'area' because it read it in as 'AREA'.

Shapefile seems to have the convention to use uppercase column names. As Shapefile with their limits in length etc are somewhat special in many cases anyway, I'd ignore that and just tell people to not use shapefiles honestly.

fiboa_cli/datasets/digifarm.py
cholmes: Their area field is in sq meters. So what I'm trying to do here is just use their hectares as area, and then use their area as just a custom field for digifarm called area_sq_m.

But I then got some data that didn't have area_ha, so I think the best approach is likely to do area: area, and then do the column transform.

m-mohr: Ah, gotcha. Sounds good. The value of having m² in addition to ha seems pretty limited to me...

cholmes: Yeah, I was trying to stay 'faithful' to theirs:

```
      "properties": {
        "id": "10149_no_202012",
        "area": 100095.056069513,
        "area_acres": "24.734",
        "area_ha": "10.010"
      },
```

I'll ask to see if they feel it is really important to have all 3.

m-mohr: area_ha as a string? And the rounding is also a bit weird... :-o

m-mohr: To clean things up I'll convert from area (m²) to hecatares and ignore the other fields.

m-mohr:
I've merged this for now for GeoJSON inputs.
We can extend it to be able to read other formats later.

PR #70, opened 2024-06-23
Fieldscapes converter for AI4SmallFarms

snehalchaudhari98:
Adding data converter for the sub-sampled data for Cambodia and Vietnam.

PR #83, opened 2024-07-25
Converters for AI4smallholders

cholmes:
Taking #70 a bit further - converting to shorter names, and calling from the source data. 

So far have just gotten 2 gpkg's from cambodia working, but it seems to work, so mostly just need to add all the file links for cambodia, and then do vietnam similarly.


m-mohr:
I've worked to complete this. I'll not make this data available on Source though.

I didn't see a good reason to have two converters, I merged them.

cholmes:
Thanks! I'll try to do the stuff to make it available on source. And yeah, no reason to have two converters - my intent was to get rid of the other PR when this was done.

PR #91, opened 2024-09-04
Croatia converter

Barasakar:
Hello @m-mohr, I have finished the converter, but it is not verified yet as the converter takes a long time to run and it gets killed in the terminal, so I am making this draft PR. 

Is it normal that a converter took a while to process the data? The data size is about 870MB (when zipped). I noticed some of your converters also deal with large dataset but they run fine. If you could give me some suggestions how to address this issue, I will greatly appreciate it!


m-mohr:
Thanks, I'm hitting the same issue, regardless of whether I generate GeoParquet 1.0 or 1.1, before and after the updates in the geopandas code. Some column filters can be removed, but I'll investigate further and update you whenever I have more details.

m-mohr:
Issues fixed, waiting for tests, then it's good to go.

PR #95, opened 2024-11-08
Alternative, freely licensed source for BE-WA

ivorbosloper:
The inspire Land Use dataset is similar and freely-available. See also https://github.com/fiboa/data-survey/pull/22


fiboa_cli/datasets/be_wa.py
```diff
@@ -60,12 +62,13 @@
...
 
 def migration(gdf):
-    gdf['crop_name'] = gdf['crop_name'].str.strip()  # .str.replace("  ", " ").str.replace("( ", "(")
-
     # Extract last digits from crop_code, example crop_code is:
     # http://geoservices.wallonie.be/inspire/atom/LU_LandUseClassification_LPIS.xml?code=#LandUseClass.lpis.cropCategory.6
     gdf['crop_code'] = gdf['crop_code'].str.extract(r'\.(\d+)$', expand=False)
+    gdf['crop_name'] = gdf['crop_name'].str.strip()  # .str.replace("  ", " ").str.replace("( ", "(")
     gdf["determination_datetime"] = "2022-01-01T00:00:00Z"
+    gdf["area"] = gdf["geometry"].area
```
m-mohr, line 70: We don't do the automatic addition of area/perimeter usually. We likely add a generic parameter for this soon, so I'd propose to remove this here.

fiboa_cli/util.py
```diff
@@ -105,6 +105,8 @@ def get_fs(url_or_path: str) -> AbstractFileSystem:
     parsed = urlparse(url_or_path)
 
     if parsed.scheme in ('http', 'https'):
+        if re.search(r"[?&]blocksize=0", url_or_path):
+            return HTTPFileSystem(block_size=0)
```
m-mohr, line 109: That looks suspicious. What's this? :-)

ivorbosloper: I agree. However the program fails without this option, and the error suggests adding this `blocksize=0` option. We're using an `fsspec.HTTPFileSystem` and download the file chunked in `convert_utils.py:stream_file()`. This is implemented by doing http-range requests of 10M size, but the origin server for this dataset does not support http-range requests and fails.

This was the shortest path to introducing this option.

https://filesystem-spec.readthedocs.io/en/latest/api.html#fsspec.implementations.http.HTTPFileSystem

fiboa_cli/datasets/be_wa.py
ivorbosloper: Will do

fiboa_cli/util.py
ivorbosloper: Yes, I tried: https://github.com/fiboa/cli/pull/95/files#diff-73f13a18826c38c2ef4d83a2a5252675d1b156fad00c4f39b3cd9b51d44e75f7R108

All source datasets slightly differ. It leads us to a `convert_utils.py:convert()` function with 25 arguments + kwargs, and it still is not enough to capture all variations. I muffled this `blocksize` argument into the url to avoid introducing more. If we would refactor everything to a class-based Converter approach (with a BaseConverter, maybe some mixins, for each country a derrived Converter class), we could override a `self.get_fs()` and handle local variations locally...

m-mohr: Yeah, we are hitting limits here ;-)

Thanks for adding the comment.

m-mohr (approved):

PR #96, opened 2024-11-08
Add Sweden

ivorbosloper:
Add Sweden from original source file.

Use ec-HCAT csv-file for code to name mapping. We can use such a setup for the hcat-extension, and to quickly map the internal codes to hcat-codes. Not sure if we should copy the csv files into this repo.


fiboa_cli/__init__.py
```diff
@@ -381,15 +383,21 @@ def jsonschema(schema, out, fiboa_version, id_):
...
     """
     Converts existing field boundary datasets to fiboa.
     """
     log(f"fiboa CLI {__version__} - Convert '{dataset}'\n", "success")
     try:
-        convert_(dataset, out, input, cache, source_coop, collection, compression, geoparquet1)
+        convert_(dataset, out, input, cache, source_coop, collection, compression, geoparquet1, mapping_file)
     except Exception as e:
-        log(e, "error")
+        log(traceback.format_exc(), "error")
```
m-mohr, line 400: Is that intentional?

I know it's annoying that the CLI doesn't provide tracebacks, but on the other hand we may not want them in production to always show up. We should try to find a generic better solution here, I think.

It would've been nice in general to keep this open for a review...

m-mohr: => #101

PR #99, opened 2024-11-10
Add Luxembourg

ivorbosloper:
Solve https://github.com/fiboa/data/issues/24

### DIFF
--- a/fiboa_cli/datasets/lu.py
+++ b/fiboa_cli/datasets/lu.py
@@ -28,16 +28,14 @@ PROVIDERS = [
     }
 ]
 ATTRIBUTION = "Luxembourg ministry of Agriculture"
-LICENSE = {
-    "title": "This layer is published under the terms of the license Creative Commons Attribution 4.0 International (CC BY 4.0)",
-    "href": "https://creativecommons.org/licenses/by/4.0/",
-    "type": "text/html",
-    "rel": "license"
-}
+LICENSE = "CC-BY-4.0"
 
+EXTENSIONS = [
+    "https://fiboa.github.io/flik-extension/v0.1.0/schema.yaml"
+]
 COLUMNS = {
     "geometry": "geometry",
-    "FLIK": "id",
+    'FLIK': ['id', 'flik'], # make flik id a dedicated column to align with NRW etc.
     "determination_datetime": "determination_datetime",
 }
 
@@ -52,6 +50,7 @@ def convert(output_file, cache = None, **kwargs):
         TITLE,
         DESCRIPTION,
         providers=PROVIDERS,
+        extensions=EXTENSIONS,
         attribution=ATTRIBUTION,
         license=LICENSE,
         **kwargs
--- a/fiboa_cli/datasets/lu.py
+++ b/fiboa_cli/datasets/lu.py
@@ -1,15 +1,8 @@
-import os
-
-from ..convert_gml import gml_assure_columns
-from ..util import log
-import geopandas
-
 from ..convert_utils import convert as convert_
 
 SOURCES = {
     "https://data.public.lu/fr/datasets/r/b4ae6690-7e4c-4454-8b60-9fa33ba6a61b": "lu.zip"
 }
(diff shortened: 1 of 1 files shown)


fiboa_cli/datasets/lu.py
```diff
@@ -0,0 +1,58 @@
...
+        "roles": ["licensor", "distributor"]
+    }
+]
+ATTRIBUTION = "Luxembourg ministry of Agriculture"
+LICENSE = {
+    "title": "This layer is published under the terms of the license Creative Commons Attribution 4.0 International (CC BY 4.0)",
+    "href": "https://creativecommons.org/licenses/by/4.0/",
+    "type": "text/html",
+    "rel": "license"
+}
```
m-mohr, line 36: This can be simplified.

```suggestion
LICENSE = "CC-BY-4.0"
```

fiboa_cli/datasets/lu.py
```diff
@@ -0,0 +1,58 @@
...
+
+COLUMNS = {
+    "geometry": "geometry",
+    "FLIK": "id",
```
m-mohr, line 40:
```suggestion
    'FLIK': ['id', 'flik'], # make flik id a dedicated column to align with NRW etc.
```

fiboa_cli/datasets/lu.py
```diff
@@ -0,0 +1,58 @@
...
+]
+ATTRIBUTION = "Luxembourg ministry of Agriculture"
+LICENSE = {
+    "title": "This layer is published under the terms of the license Creative Commons Attribution 4.0 International (CC BY 4.0)",
+    "href": "https://creativecommons.org/licenses/by/4.0/",
+    "type": "text/html",
+    "rel": "license"
+}
+
+COLUMNS = {
```
m-mohr, line 38:
```suggestion
EXTENSIONS = [
    "https://fiboa.github.io/flik-extension/v0.1.0/schema.yaml"
]
COLUMNS = {
```

fiboa_cli/datasets/lu.py
```diff
@@ -0,0 +1,58 @@
...
+    convert_(
+        output_file,
+        cache,
+        SOURCES,
+        COLUMNS,
+        ID,
+        TITLE,
+        DESCRIPTION,
+        providers=PROVIDERS,
+        attribution=ATTRIBUTION,
```
m-mohr, line 55:
```suggestion
        extensions=EXTENSIONS,
        attribution=ATTRIBUTION,
```

ivorbosloper (commented):
Thanks, good suggestion @m-mohr

fiboa_cli/datasets/lu.py
```diff
@@ -0,0 +1,57 @@
+import os
+
+from ..convert_gml import gml_assure_columns
+from ..util import log
+import geopandas
+
+from ..convert_utils import convert as convert_
```
m-mohr, line 7:
```suggestion
from ..convert_utils import convert as convert_
```

None of the imports is actually used, right?

m-mohr (approved):

fiboa_cli/datasets/lu.py
ivorbosloper: Good point. I currently don't run an import-optimizer. If we could pick a code style with a deterministic code formatter (ruff, black, pylint) we could run that automatically before every commit (with a git hook) and check code style with github action. Just fixed it now... :)

ivorbosloper:
Hmmm. Too soon I'm afraid. I tried re-generating + publishing the Luxembourg files for source.coop to include the FLIK attribute, but the FLIK format is not the same. Should we revert or should the FLIK spec be more flexible?

```
Validating fiboa/data/lu/luxembourg.parquet
  - fiboa version: 0.2.0
    fiboa extensions:
      - https://fiboa.github.io/flik-extension/v0.1.0/schema.yaml
  - flik: String '0170586' is shorter than the minimum length of 16.
  - flik: String '0170586' does not match the required pattern: ^[A-Z]{2}[A-Z0-9]{2}[A-Z0-9]{2}[A-Z0-9]{10}$.
  ```

m-mohr:
That's not a valid FLIK ID. Maybe they removed the prefix? Let me check... 🤔

Issue #100, opened 2024-11-11
Refactor converters into class-based modules


m-mohr:
This would help get rid of the growing number of parameters and should make implementations a bit more simple.

Note: We'd need to update the tutorials and videos for creating converters.

m-mohr:
Seems some converters are left, so reopening

PR #101, opened 2024-11-11
Refactoring

m-mohr:
Refactoring for consistency, add index_as_id option as it seems to become more common.

After the workshop I'll try to implement #100 so that we get rid of all the arguements ;-)


ivorbosloper (approved):
Looks good

Issue #90, opened 2024-08-24
`merge`: Combine datasets


m-mohr:
@cholmes wrote:

> Fiboa ‘merge’ command (I think I still don’t have a ticket for this). Should have a command that makes it easy for people to combine data. There’s some of this in support of FTW data release, and that’s likely the ‘easy’ case, where none of the boundaries overlap. But it’d be good to have a way to merge multiple years, like from one of the government data sets, where there’s a ‘latest’ that gets used, and then also the history (this obviously needs a good bit of work to determine the right fiboa extensions, but we could start with trying to merge existing data). And then also to do a merge of two datasets that have overlaps, and have one or two strategies for how to pick which one if there are conflicts. Could do some of those as ‘operations’ - merge is simple, and then people can use operations to clean it up.

andyjenkinson:
What you're describing is extremely complex because overlaps are not 1:1, there are chains of consequences. For example:
* A overlaps B
* A overlaps C
* A is newer than B
* C is newer than A

Also even for the simple cases where you have an entire dataset which are all 'newer', what you find is that often a 'newer' set of boundaries don't overlap with each other, but it's not possible to introduce them 1-by-1 because the intermediate state introduces overlaps. Basically for both of these reasons it requires set-level comparison logic. We approach this in stages by computing non-overlapping 'islands', then processing each in chunks. This still also requires that the user is happy with a simple logic of 'is newer' - in fact often it is the case that it's undesirable to replace an older, higher quality one with a newer predicted boundary.

The other complication by the way is that even if boundaries don't overlap, identifiers are only unique within a single dataset. Again that's why we create a separate Feature object with a (different) guaranteed-unique Boundary ID and then link each 'input' object's metadata as a separate object in order to retain the original metadata. (In GFID this has to be a 1:many relationship because we deduplicate the geometries, and even when two objects have the same geometry they typically have different (conflicting) metadata - e.g. two different dates, two different claimed delineation methods. We do this because it's actually quite common to see the same geometry appear in multiple datasets, e.g. FotW contains many national LPIS geometries, each which itself is a separately published dataset.

This is why solving the deduplication problem required creating an entire product (i.e. Global FieldID) focused on that goal. Ideally we would just make it so that  also why we cannot make GFID fiboa-compliant, because the FIBOA specification is not compatible

m-mohr:
[...]

PR #102, opened 2024-11-11
Add simple merge command #90

m-mohr:
Implements #90


tests/test_merge.py
```diff
@@ -0,0 +1,34 @@
...
+    result = runner.invoke(merge, args)
+    assert result.exit_code == 0, result.output
+
+    # Merged parquet file should be valid
+    result = runner.invoke(validate, [tmp_file.name, '--data'])
+    assert result.exit_code == 0, result.output
+
+    parquet = import_optional_dependency(
+        "pyarrow.parquet", extra="pyarrow is required for Parquet support."
+    )
```
m-mohr, line 28: Is that an actual issue? We use it in other places, too, so it's not really optional, isn't it?

ivorbosloper: I searched in the code on how to open/read/write parquet files (for testing this case). This is what I found in `geopandas.py` and I bluntly assumed this is how I should import it... Feel free to change

ivorbosloper (approved):
Tested, it works

m-mohr:
Is there a better default?

PR #103, opened 2024-11-11
Remove FLIK from LU

m-mohr:
As discussed, I think...


ivorbosloper:
Yes, FLIK stands for "Flächen-Identifikator" (some kind of field block id) and is interpreted differently in each member state...

PR #108, opened 2024-11-12
Change Parquet row group size

m-mohr:
The default row group size of exported parquet files was changed from ~1.000.000 to 25.000


cholmes (approved):

ivorbosloper (approved):
Good point

PR #111, opened 2024-11-13
Lithuania converter

cholmes:
Took over #69 and fleshed it out a bit more.

It's not quite working right yet, as I couldn't get it to remove z-coordinates, which causes trouble for the geoparquet writer. I tried to add some code for shapely to get rid of them, but when I run it says '/Users/cholmes/miniforge3/envs/fiboa/lib/python3.12/site-packages/pyogrio/raw.py:198: RuntimeWarning: driver ESRI Shapefile does not support open option MIGRATIONS' and it doesn't work. 

I got things working by just hand-editing the source input file, so was able to debug the next problem I had (casting properly from string to uint32).

I made it so the output filters in the same way that FTW does, but I have an open question for them about rapeseed, which is filtered out but seems like it should be in the dataset? Once I clarify I'll adjust if needed.

It does seem like the source data isn't spatially sorted at all. 

I also need to update this to main, as it's still making geoparquet 1.0. I had troubles doing that on my local environment, so going to try on github.


cholmes:
Tried again merging main in locally and got it updated, but I seem unable to figure out how to update the migrate function to work. I'm getting 'convert() got an unexpected keyword argument 'store_collection'' - I tried to remove it but seem to still get the same error (but may be some weird build thing).

Anyways, I think this one is pretty close, but I'd need some help or at least some tips to take it over the line.

m-mohr:
I'll have a look

m-mohr:
@cholmes 4b49e0b60bc91c07d5435214d4a3fde3789204c4 makes it work, please have a look :-)

m-mohr (approved):

PR #118, opened 2024-11-14
Converter for Slovenia crop fields

ivorbosloper:


fiboa_cli/convert_utils.py
```diff
@@ -402,6 +402,10 @@ def download_files(uris, cache_folder = None):
             elif py7zr.is_7zfile(cache_file):
                 with py7zr.SevenZipFile(cache_file, 'r') as sz_file:
                     sz_file.extractall(zip_folder)
+            elif name.endswith(".rar"):
+                import rarfile
+                with rarfile.RarFile(cache_file, 'r') as w:
+                    w.extractall(zip_folder)
```
m-mohr, line 408: If we add the rar extract directly here, I think the dependency should just be in the normal dependencies, not in the converter extras.

fiboa_cli/datasets/si.py
```diff
@@ -0,0 +1,80 @@
...
+around 150 different crop categories.
+"""
+PROVIDERS = [
+    {
+        "name": "Ministry of Agriculture, Forestry and Food (Ministrstvo za kmetijstvo, gozdarstvo in prehrano)",
+        "url": "https://www.gov.si/drzavni-organi/ministrstva/ministrstvo-za-kmetijstvo-gozdarstvo-in-prehrano/",
+        "roles": ["producer", "licensor"]
+    }
+]
+ATTRIBUTION = ""
```
m-mohr, line 23: Can be removed if empty

fiboa_cli/convert_utils.py
ivorbosloper: Ah. Yes. My line of thinking is that the rarfile library requires some command-line tools to be installed. https://pypi.org/project/rarfile/ . If the command-line availability is already checked during `pip install` (of which I'm not sure), we force the availability of unrar for every use of fiboa.

> Compressed files are extracted by executing external tool: [unrar](https://www.rarlab.com/) (preferred), [unar](https://theunarchiver.com/command-line), [7zip](https://www.7-zip.org/) or [bsdtar](https://github.com/libarchive/libarchive).

m-mohr: Oh, I wasn't aware. Had hoped this is a pure Python library.

It didn't check the availability of the external tools during the install of rartools for me, it just completed and I have none of the tools installed.
All dependencies in the "core" should be part of the dependencies, otherwise the rar code should probably move to the converter file itself.

m-mohr (approved):

m-mohr:
Please add a changelog entry and then you can merge...

PR #120, opened 2024-11-14
Extend Croatia with crop mapping

ivorbosloper:


m-mohr:
A changelog please :-)

fiboa_cli/datasets/hr.py
```diff
@@ -141,12 +141,47 @@
...
+321;Krški pašnjak;Karst pasture
+410;Vinograd;Vineyard
+411;Iskrčeni vinograd;Cleared vineyard
+421;Maslinik;Olive grove
+422;Voćnjak	Poljoprivredno;Orchard
+430;Kulture kratke ophodnje;Short tour cultures
+450;Rasadnik;Nursery
+490;Mješoviti višegodišnji nasadi;Mixed perennial plantations
+900;Ostale vrste uporabe zemljišta;Other types of land use
+910;Privremeno neodržavana parcela;Temporarily unmaintained plot"""
```
m-mohr, line 172: I'm wondering whether we should store these in separate files...

ivorbosloper: We could store it as a separate file. I made a proposal

fiboa_cli/datasets/commons/data.py
```diff
@@ -0,0 +1,6 @@
+from csv import DictReader
+from os.path import dirname, join
+
+
+def read_data_csv(name, **kwargs):
+    return list(DictReader(open(join(dirname(dirname(__file__)), "data-files", name), 'r'), **kwargs))
```
m-mohr, line 6:
```suggestion
    path = join(dirname(dirname(__file__)), "data-files", name)
    with open(path, 'r') as f:
        return list(DictReader(f, **kwargs))
```

m-mohr (approved):
A small code suggestion, otherwise LGTM

fiboa_cli/datasets/hr.py
m-mohr: Thanks

PR #127, opened 2024-11-21
GIve a clear error if column names become duplicated in the dataframe

ivorbosloper:
Assists debugging of converters

### DIFF
--- a/fiboa_cli/parquet.py
+++ b/fiboa_cli/parquet.py
@@ -43,6 +43,11 @@ def create_parquet(data, columns, collection, output_file, config, missing_schem
     # Update the GeoDataFrame with the correct types etc.
     data = update_dataframe(data, columns, schemas)
 
+    _columns = list(data.columns)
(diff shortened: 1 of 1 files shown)


fiboa_cli/parquet.py
```diff
@@ -43,6 +43,11 @@ def create_parquet(data, columns, collection, output_file, config, missing_schem
     # Update the GeoDataFrame with the correct types etc.
     data = update_dataframe(data, columns, schemas)
 
+    _columns = list(data.columns)
+    duplicates = set([x for x in _columns if _columns.count(x) > 1])
+    if len(duplicates):
+        raise ValueError(f"Columns are not unique: {duplicates}")
```
m-mohr, line 49:
```suggestion
        raise ValueError(f"Columns are defined multiple times: {duplicates}")
```

"Columns are not unique" sounds a bit like the values in the columns are not unique, but I guess that's not what is meant...

m-mohr (approved):
Just some textual changes would be nice, otherwise LGTM

PR #129, opened 2024-11-27
Stay consistent with iso-code naming scheme

ivorbosloper:
It's WAL according to https://en.wikipedia.org/wiki/ISO_3166-2:BE#Regions

### DIFF
--- a/fiboa_cli/datasets/be_wal.py
+++ b/fiboa_cli/datasets/be_wal.py
@@ -7,7 +7,7 @@ SOURCES = {
 }
 LAYER = "ExistingLandUseObject"
 
-ID = "be_wa"
+ID = "be_wal"
 SHORT_NAME = "Belgium, Wallonia"
 TITLE = "Belgium Wallonia: Parcellaire Agricole Anonyme"
 DESCRIPTION = """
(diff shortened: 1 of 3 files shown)


m-mohr:
There are more converters that don't follow the ISO-3166-2 codes, e.g. Germany. Should we change that generally and does this have any implications for the deployed files? Thoughts?

Note: The codes I used for Germany are more common in Germany (NRW vs NW, NDS vs NI) itself that's why I chose them initially, but they are also not really standardized in any form...

ivorbosloper:
> There are more converters that don't follow the ISO-3166-2 codes, e.g. Germany. Should we change that generally and does this have any implications for the deployed files? Thoughts?
> 
> Note: The codes I used for Germany are more common in Germany (NRW vs NW, NDS vs NI) itself that's why I chose them initially, but they are also not really standardized in any form...

I'm not sure. Level-1 ISO-country codes really make sense. The level-2 are sometimes 2 characters, sometimes 3 (depends per country), and I'm not so familiar with them. I prefer to stick with some format within a single country, and having BE-WA and BE-VLG seems inconsistent...

m-mohr (approved):

PR #132, opened 2024-12-03
Spain, each autonomous region it's own converter

ivorbosloper:
The different sources have different licenses/structure


m-mohr (dismissed):
This is based on #138 so we need to merge that PR first...

m-mohr:
This needs a rebase :-)

ivorbosloper:
@m-mohr there is another relevant issue here too: I started to use code lists on fiboa.org, e.g. https://fiboa.org/code/es/sigpac/land_use.csv . Previously it was  https://github.com/fiboa/cli/blob/main/fiboa_cli/datasets/data-files/es_coda_uso.csv or the discussed downloadable https://raw.githubusercontent.com/fiboa/cli/refs/heads/main/fiboa_cli/datasets/data-files/es_coda_uso.csv . I like it a lot better but please review this strategy...

m-mohr:
The approach is fine, I think. It makes us less dependant on the GitHub URL and is nicer to read.

Both approaches were not really versioned yet. I'm wondering whether we should introduce a version into the URL in case changes are made? e.g. https://fiboa.org/code/es/sigpac/v1/land_use.csv
I think we don't need version vers often, usually new lines are added to the lists over time, but I could see it coming that some list change in a way that old results are not reproducible any longer or even worse the whole list changes drastically or parts are removed.

m-mohr:
Okay, that also works for me. Just doesn't look as nice ;-)

fiboa_cli/convert.py
```diff
@@ -89,7 +89,7 @@ def read_converter(_id):
     if not hasattr(module, "convert"):
         try:
             clazz = next(v for v in module.__dict__.values()
-                         if type(v) is type and issubclass(v, BaseConverter) and not v.__name__.startswith("Base"))
+                         if type(v) is type and issubclass(v, BaseConverter) and "BaseConverter" not in v.__name__)
```
m-mohr, line 92: Is that intentional?

fiboa_cli/convert_utils.py
```diff
@@ -545,3 +558,107 @@ def hash_df(df):
     buf = StringIO()
     df.info(buf=buf)
     return hash(buf.getvalue())
+
+
+class EsriRESTPregenerateConverterMixin:
```
m-mohr, line 563: Can we put the two classes into separate files?

ivorbosloper: Yes

fiboa_cli/convert_utils.py
```diff
@@ -120,7 +122,7 @@ def normalize_geojson_properties(feature):
     return feature
 
 
-def read_geojson(path, **kwargs):
+def read_geojson(path, layer=None, **kwargs):
```
m-mohr, line 125: The layer seems to be unused. What's the purpose?

fiboa_cli/convert_utils.py
```diff
@@ -246,11 +251,11 @@ def download_files(self, uris, cache_folder=None):
                 elif py7zr.is_7zfile(cache_file):
                     with py7zr.SevenZipFile(cache_file, 'r') as sz_file:
                         sz_file.extractall(zip_folder)
-                elif name.endswith(".rar"):
+                elif rarfile.is_rarfile(cache_file):
```
m-mohr, line 254: I think there was an issue with this method for the SI converter, but not sure. As there's no test for SI, can you check whether the SI converter still works fine with this?

fiboa_cli/convert_utils.py
```diff
@@ -209,9 +214,9 @@ def download_files(self, uris, cache_folder=None):
                     name = name_from_uri(uri)
                     # if there's no file extension, it's likely a folder, which may not be unique
                     if "." not in name:
-                        name = str(i)
```
m-mohr, line 212: If i is not used anymore, we can remove it above.

fiboa_cli/convert_utils.py
```diff
@@ -276,6 +281,10 @@ def get_urls(self):
     def read_data(self, paths, **kwargs):
         gdfs = []
         for path, uri in paths:
+            if "*" in path:
+                lst = glob(path)
+                assert len(lst) == 1, f"Can not match {path} to a single file"
+                path = lst[0]
```
m-mohr, line 287: A comment of the purpose would be helpful, I think. It seems this is for a specific usecase...

Also, glob is a bit tricky on Windows. Works probably in 99% of the cases, but if there's a better solution, we should aim for that ;-)

fiboa_cli/convert_utils.py
```diff
@@ -432,28 +441,29 @@ def convert(self, output_file, cache=None, input_files=None, source_coop_url=Non
         if directory:
             os.makedirs(directory, exist_ok=True)
 
-        urls = self.get_urls()
         if input_files is not None and isinstance(input_files, dict) and len(input_files) > 0:
             log("Using user provided input file(s) instead of the pre-defined file(s)", "warning")
             urls = input_files
-        elif urls is None:
+        elif (urls := self.get_urls(cache=cache)) is None:
```
m-mohr, line 447: It seems the only thing that is ever passed is cache, but the methods accepts kwargs. So to understand what's actually passed in here, you need to look through the code. Wouldn't it make more sense to use cache instead of kwards in those get_urls implementations?

fiboa_cli/datasets/es_nc.py
```diff
@@ -0,0 +1,68 @@
...
+        "crop:name": "crop:name",
+        "crop:name_en": "crop:name_en",
+    }
+    column_migrations = {
+        "BEGINLIFE": lambda col: pd.to_datetime(col, format="%d/%m/%Y"),
+    }
+    use_code_attribute = "IDUSO24"
+    index_as_id = True
+
+    def get_urls(self, cache=None):
```
m-mohr, line 42: A method called get_ that is downloading and writing something might not be ideal. Why is this done here?

fiboa_cli/converter_rest.py
```diff
@@ -0,0 +1,85 @@
...
+
+        base_url = paths[0]  # loop over paths to support more than 1 source
+
+        source_fs = get_fs(base_url)
+        cache_fs = None
+
+        if self.cache_folder:
+            cache_fs = get_fs(self.cache_folder)
+            if not cache_fs.exists(self.cache_folder):
+                cache_fs.makedirs(self.cache_folder)
```
m-mohr, line 47: I think this code is copied from the convert utils. Can we make this a function so that we deduplicate the code a bit?

fiboa_cli/converter_rest.py
```diff
@@ -0,0 +1,85 @@
...
+            data = gpd.read_file(url)
+            print(f"Read {len(data)} features, page {len(gdfs)} from [{data.iloc[0, 0]} ... {data.iloc[-1, 0]}]")
+            last_id = data[self.rest_attribute].values[-1]
+
+            # 0. Run migration per file/layer
+            data = self.file_migration(data, base_url, base_url, layer["id"])
+            if not isinstance(data, gpd.GeoDataFrame):
+                raise ValueError("Per-file/layer migration function must return a GeoDataFrame")
+
+            gdfs.append(data)
```
m-mohr, line 82: Same here, ideally deduplicate code.

fiboa_cli/convert_utils.py
```diff
@@ -482,13 +492,14 @@ def convert(self, output_file, cache=None, input_files=None, source_coop_url=Non
...
             gdf = gdf.explode()
+            gdf = gdf[np.logical_and(gdf.geometry.type == "Polygon", gdf.geometry.is_valid)]
             if gdf.geometry.array.has_z.any():
                 log("Removing Z geometry dimension", "info")
                 gdf.geometry = gdf.geometry.force_2d()
 
         gdf = self.post_migrate(gdf)
 
-        if hash_before != hash_df(gdf):
+        if hash_before != hash_df(gdf.head()):
```
m-mohr, line 502: How fast is this hashing? We could potentially also just always print the head here...

fiboa_cli/convert_utils.py
```diff
@@ -482,13 +492,14 @@ def convert(self, output_file, cache=None, input_files=None, source_coop_url=Non
         if not original_geometries:
             gdf.geometry = gdf.geometry.make_valid()
             gdf = gdf.explode()
+            gdf = gdf[np.logical_and(gdf.geometry.type == "Polygon", gdf.geometry.is_valid)]
```
m-mohr, line 495: MultiPolygons never happen due to the explode, right?

fiboa_cli/convert_utils.py
ivorbosloper: The [read_geojson() invocation](https://github.com/fiboa/cli/blob/spain/fiboa_cli/convert_utils.py#L307) is passed a `layer` keyword-arg, due to [the optional layer filtering feature](https://github.com/fiboa/cli/blob/spain/fiboa_cli/convert_utils.py#L292). This layer param avoids passing this `layer` kwarg to the [open() call in read_geojson()](https://github.com/fiboa/cli/blob/spain/fiboa_cli/convert_utils.py#L293). Alternatively, we could avoid filtering for layers in both parquet & geojson; I think this layer concept is not supported in geojson either.

fiboa_cli/convert_utils.py
ivorbosloper: Will add test for SI...

fiboa_cli/convert_utils.py
ivorbosloper: Valid point...

fiboa_cli/convert_utils.py
m-mohr: Yeah, layers is not really a thing in GeoJSON. Not even sure what that means. Does this filter on a layer property or so?
Rhis feels like a case where it should not be supported by the generic reader but be somehow implemented only in the specific converter by overriding a portion of the reader or so.

fiboa_cli/convert_utils.py
ivorbosloper: Will add comment. The challenge is that some converter-sources are comprised of many different zip files, and the shapefile naming scheme in these zip files is not consistent/predictable. This will allow a `*.shp` to just match any shapefile... Unpacking/inspecting the zip files in the converter.py file duplicates logic (but keeps the main-file simpler).

m-mohr: I guess we can keep it and try it out until we run into issues ;-)

fiboa_cli/convert_utils.py
ivorbosloper: Even cache is not used. I made this overridable "hook"-function overly generic so we could later add new parameters without changing the implementations. But YAGNI and let's get rid of it all...

fiboa_cli/convert_utils.py
ivorbosloper: Hashing is fast because we only use head. Now we only print again if something has changed (makes debugging easier).

fiboa_cli/converter_rest.py
ivorbosloper: We could move the part that actually get the data based on paths+kwargs to some iterator/function returning dataframes, like `self.get_data(paths, **kwargs) --> iterator[DataFrame]`. I'll fix that.

fiboa_cli/convert_utils.py
ivorbosloper: Fixed "assignment in expression"

fiboa_cli/convert_utils.py
ivorbosloper: No, ogr supports layers (because shapefile and gpkg support combining layers). But geojson files can't embed this, and GEOS makes a 'dummy' layer with the name of the file...

fiboa_cli/convert_utils.py
ivorbosloper: Yes, let me know if it fails on windows...

fiboa_cli/datasets/es_nc.py
ivorbosloper: Valid point, separating will make the code more readable...

CHANGELOG.md
```diff
@@ -10,6 +10,21 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
...
+- Converter for Spain - Catalonia
+- Converter for Spain - Cantabria
+- Converter for Spain - Castilla y León
+- Converter for Spain - Castilla-La Mancha
+- Converter for Spain - Canary Islands
+- Converter for Spain - Extremadura
+- Converter for Spain - Galicia
+- Converter for Spain - Madrid
+- Converter for Spain - Navarra
+- Converter for Spain - Valencia
```
m-mohr, line 26:
```suggestion
- Converters for Spain: Aragon, Andalusia, Balearic Islands, Basque Country, Catalonia, Cantabria, Castilla y León, Castilla-La Mancha, Canary Islands, Extremadura, Galicia, Madrid, Navarra, Valencia
```

m-mohr (approved):

PR #133, opened 2024-12-03
Implement LV based on paged WFS reading

ivorbosloper:
Parse data from the original source (LV Government). This will allow us to add more years to the data collection (EuroCrops now only has 2021).


fiboa_cli/convert_utils.py
```diff
@@ -87,7 +87,14 @@ def convert(
             elif is_json:
                 data = read_geojson(path, **kwargs)
             else:
-                data = gpd.read_file(path, **kwargs)
+                try:
+                    data = gpd.read_file(path, **kwargs)
+                except IndexError as e:
+                    if str(e).startswith("index 0 is out of bounds") and len(paths) > 1:
```
m-mohr, line 93:
```suggestion
                    # Error may occur due to paged reading of APIs, e.g. in the LV converter
                    if str(e).startswith("index 0 is out of bounds") and len(paths) > 1:
```

fiboa_cli/datasets/lv.py
```diff
@@ -0,0 +1,86 @@
...
+    }
+}
+
+
+def convert(output_file, cache = None, mapping_file = None, **kwargs):
+    count = 3000
+    sources = {
+        f"{SOURCES}?request=GetFeature&service=wfs&version=2.0.0&typeNames=Lauki&count={count}&startindex={count * i}": f"lv_{i}_{count}.xml"
+        for i in range(500000 // count)  # TODO number should be dynamic, stop reading with 0 results
+    }
```
m-mohr, line 58: Do you want to solve the TODO or do it later?

Generally, I think we should open an issue to implement paged WFS / OGC API - Features reading generically. This will likely occur in more converters.

m-mohr (changes requested):
Changelog :-)

fiboa_cli/datasets/lv.py
ivorbosloper: Yes, this is a bit ugly but easier to solve it later. Currently, there is no mechanism to override the sourcing of data. Will add it in class-based Converter approach. `PagedWFSReader` can become a mixin that overrides the method for getting the features.

ivorbosloper:
> Changelog :-)
> 
> I'm inclined to create a PR template that asks to create a changelog entry ;-)

I'd love that! For every PR that changes more than say 10 lines; `changelog.md` should contain a some change :)

m-mohr:
I was more thinking of a checkbox in the PR description fields (similar to issue templates), but I guess a CI check also works (although a one line change could also be significant) ;-)

m-mohr (approved):

PR #135, opened 2024-12-05
Japan based on converted parquet files, rewrite to source-based parser later

ivorbosloper:


fiboa_cli/convert_utils.py
```diff
@@ -42,13 +42,18 @@ def convert(
...
         **kwargs):
     """
     Converts a field boundary datasets to fiboa.
     """
     if bbox is not None and len(bbox) != 4:
         raise ValueError("If provided, the bounding box must consist of 4 numbers")
 
+    if open_options:
+        kwargs.update(open_options)
+
```
m-mohr, line 56: Shouldn't that be possible in the converter code by updating kwargs there? So ideally we don't need another parameter...

fiboa_cli/datasets/jp.py
```diff
@@ -0,0 +1,79 @@
...
+        },
+        'admin_local_code': {
+            'type': 'string'
+        },
+    }
+}
+def migrate(df):
+    df.rename(columns={"GEOM": "geometry"}, inplace=True)
+    df.set_geometry("geometry", inplace=True)
+    return df
```
m-mohr, line 58: Hmm, ideally we could do that in the COLUMNS mapping...

fiboa_cli/convert_utils.py
ivorbosloper: Yes, but sadly my keyword-argument for the gdf-open function overlaps with the arguments of convert. I'm trying to pass `columns=["col1", "col2"]` so only a selection is loaded into the not-lazy data frame. That way I can keep the memory usage down (it now becomes 'only' ~65GB).

ivorbosloper: This pandas dataframe in-memory transformation has it's limits. I've worked with 'lazy' dataframes, I think it was with polars instead of pandas. No straightforward solution to this problem I guess https://pandas.pydata.org/docs/user_guide/scale.html ...

m-mohr: Yeah, my comment was about the additional parameters, this would get less bad with the class-based approach.

The memory issue is indeed yet another topic that will occur more and more with larger datasets.

fiboa_cli/datasets/jp.py
```diff
@@ -0,0 +1,74 @@
...
+        COLUMNS,
+        ID,
+        TITLE,
+        DESCRIPTION,
+        missing_schemas=MISSING_SCHEMAS,
+        column_migrations=COLUMN_MIGRATIONS,
+        providers=PROVIDERS,
+        attribution=ATTRIBUTION,
+        license=LICENSE,
+        open_options={"columns": ["GEOM", "polygon_uuid", "land_type_en", "local_government_cd", "issue_year"]},
```
m-mohr, line 72:
```suggestion
        # todo: See https://github.com/fiboa/cli/pull/135/files#r1872310136
        open_options={"columns": ["GEOM", "polygon_uuid", "land_type_en", "local_government_cd", "issue_year"]},
```

Just so that we know why this is the way it is.

fiboa_cli/convert_utils.py
m-mohr: Just wondering, but I think we could just rename our parameter to column_map or so. We never pass it by name, so a rename should only really have an effect in the convert_utils, not in the converters. That would make the conflict go away, which could generally be useful @ivorbosloper

ivorbosloper: > I think we could just rename our parameter to column_map or so. We never pass it by name.

Ah, hadn't seen this. Would keep it a bit simpler I suppose. Will rewrite it.

fiboa_cli/datasets/jp.py
```diff
@@ -0,0 +1,74 @@
...
+        },
+        'admin_local_code': {
+            'type': 'string'
+        },
+    }
+}
+
+
+def convert(output_file, cache = None, **kwargs):
+    sources = {YEARS[-1]: YEARS[-1].rsplit("/", 1)[-1]}
```
m-mohr, line 58: What is this doing?

fiboa_cli/datasets/jp.py
```diff
@@ -0,0 +1,74 @@
+from ..convert_utils import convert as convert_
+import pandas as pd
+
+YEARS = [
+    f"https://data.source.coop/pacificspatial/field-polygon-jp/parquet/jp_field_polygons_{year}.parquet"
+    for year in range(2021, 2024+1)]
+SOURCES = YEARS[-1]
```
m-mohr, line 7: Similarly here, why are there to variables? The whole flow of generating the sources looks confusing to me without further context.

m-mohr (changes requested):
Changelog :-)

fiboa_cli/datasets/jp.py
ivorbosloper: I'll make it simple

m-mohr (approved):
Still took me a bit to understand what's going on, but makes sense. Thanks.

PR #137, opened 2024-12-09
Require changelog for non-trivial PRs

ivorbosloper:
Test if this works


ivorbosloper:
Follow up of https://github.com/fiboa/cli/pull/133#issuecomment-2528268820

@m-mohr I just had to try it . It works. Not sure how well we like the changelog-requirement for 20-lines diff in the future

m-mohr:
I mean it's not really a requirement, and for cases where it shouldn't apply I can bypass it :-)

m-mohr:
I'd even say, if you could restrict it to only check the [fiboa_cli](https://github.com/fiboa/cli/tree/main/fiboa_cli) folder (i.e. ignore, tests, docs, CI, etc) we could even "require" it for 1+ changed lines or so.

ivorbosloper:
> I'd even say, if you could restrict it to only check the [fiboa_cli](https://github.com/fiboa/cli/tree/main/fiboa_cli) folder (i.e. ignore, tests, docs, CI, etc) we could even "require" it for 1+ changed lines or so.

I've changed the check for the `fiboa_cli` dir. The `20` is already a bit inflated as the diff-format adds like 5 lines for every changed file. We could test how well we like this action, if it's annoying we can disable it again.

PR #138, opened 2024-12-11
Class based converter

ivorbosloper:
Class based converters. Tried to keep the change minimal. Also keep supporting the current module / function-based converters, so we can slowly rewrite if we like it.

Used the NL-converter as an example. The big advantage will be that we can use composition, override small methods that do specific things. We will add more hooks or decompose the `BaseConverter.convert()` method, so implementing classes can hook-in every step of the process.


m-mohr:
Thanks, I'll have a look. May take a couple of days though.

fiboa_cli/__init__.py
```diff
@@ -343,6 +343,7 @@ def jsonschema(schema, out, fiboa_version, id_):
 ## CONVERT
 @click.command()
 @click.argument('dataset', nargs=1, type=click.Choice(list_all_converter_ids()))
+@click.argument('variant', required=False)
```
m-mohr, line 346: What's variant doing? It should have som documentation.
I think if it's optional I'd prefer it being an option instead of an argument. If it's for choosing a year, should we just call it time/year/... or so?

m-mohr:
I think I'll do a release for fiboa CLI before we merge this, just so that we have all changes released before we do this larger rework. Then we don't need to rush this.

fiboa_cli/convert.py
```diff
@@ -53,7 +57,10 @@ def list_all_converters(keys):
         try:
             converter = read_converter(id)
             for key in keys:
-                value = getattr(converter, key, None)
+                if isinstance(converter, BaseConverter):
+                    value = getattr(converter, key.lower())
+                else:
+                    value = getattr(converter, key, None)
```
m-mohr, line 63: I think we should add comments to the parts that can be refactored/simplified once we reimplemented all converters to the class based approach.

```suggestion
                # todo: remove conditional once class-based converters have been fully implemented
                if isinstance(converter, BaseConverter):
                    value = getattr(converter, key.lower())
                else:
                    value = getattr(converter, key, None)
```

fiboa_cli/convert.py
```diff
@@ -69,6 +76,15 @@ def list_all_converters(keys):
...
+    module_name = f".datasets.{_id}"
+    module = importlib.import_module(module_name, package="fiboa_cli")
+    if not hasattr(module, "convert"):
+        try:
+            clazz = next(v for v in module.__dict__.values()
+                         if type(v) is type and issubclass(v, BaseConverter) and v != BaseConverter)
+            return clazz()
+        except StopIteration:
+            log("Missing convert function or Converter class for module {_id}", "warning")
+    return module
```
m-mohr, line 90: I think we should add comments to the parts that can be refactored/simplified once we reimplemented all converters to the class based approach.

tests/conftest.py
```diff
@@ -1,8 +1,34 @@
...
 from pytest import fixture
 
+from fiboa_cli import convert_utils, util
+
 
 @fixture
 def tmp_file():
     with tempfile.NamedTemporaryFile() as out:
         yield out
+
```
m-mohr, line 11: Are the changes in the tests related to the class-based approach? If not, I think it would simplify this PR if we'd split the test changes into a separate PR.

fiboa_cli/datasets/template.py
```diff
@@ -4,170 +4,143 @@
 # The name of the file will be the name of the converter in the cli.
 # If you name it 'de_abc' you'll be able to run `fiboa convert de_abc` in the cli.
 
-from ..convert_utils import convert as convert_
+from ..convert_utils import BaseConverter
+
+
+class TemplateConverter(BaseConverter):
```
m-mohr, line 10: I think we should add a comment that explains that the class name also needs to be changed. Do we set any rules for this? e.g. CamelCase of the file name + "Converter"?

fiboa_cli/datasets/template.py
```diff
@@ -4,170 +4,143 @@
...
+    #   "https://fiboa.example/north_america.zip": ["us.gpkg", "canaga.gpkg"]
+    # }
+
+    # If multiple variants are available (e.g. different years), you can replace sources by source_variants.
+    # The dict-key can be used on the cli command line, the value will be used as 'sources'
+    #
+    # source_variants = {
+    #    "2023": "https://fiboa.example/file_2023.xyz"
+    #    "2024": "https://fiboa.example/file_2023.xyz"
+    # }
```
m-mohr, line 35: I think it could make sense to put the sources into a method which gets passed the variants argument.
Then you can create the sources more easily and don't need to repeat URIs as much...

Like: `return f"https://fiboa.example/file_{variant}.xyz"` or so. You could even do some checks, e.g. check that variant is > 2020 or so when data is only available since 2020.

fiboa_cli/datasets/template.py
```diff
@@ -4,170 +4,143 @@
...
+    license = "CC-BY-4.0"
+    # 2. a STAC Link Object with relation type "license"
+    # license = {"title": "CC-BY-4.0", "href": "https://creativecommons.org/licenses/by/4.0/", "type": "text/html", "rel": "license"}
+
+    # Map original column names to fiboa property names
+    # You also need to list any column that you may have added in the MIGRATION function (see below).
+    # GeoJSON: Nested objects can be accessed using a dot, e.g. "area.value" for {"area": {"value": 123}}
+    columns = {
+        "area_m": "area"
+    }
```
m-mohr, line 77: I assume this will need changes in the future based on the variant (e.g. a column is missing or so in the source). How do we want to handle this?

fiboa_cli/datasets/template.py
```diff
@@ -4,170 +4,143 @@
...
     10. Write the data to the Parquet file
-
-    Parameters:
-    output_file (str): Path where the Parquet file shall be stored.
-    cache (str): Path to a cached folder for the data. Default: None.
-                      Can be used to avoid repetitive downloads from the original data source.
-    source_coop_url (str): URL to the (future) Source Cooperative repository. Default: None
-    collection (bool): Additionally, store the collection separate from Parquet file. Default: False
-    compression (str): Compression method for the Parquet file. Default: brotli
-    kwargs: Additional keyword arguments for GeoPanda's read_file() or read_parquet() function.
```
m-mohr, line 149: We should probably preserve the documentation somewhere if not done yet...

fiboa_cli/convert_utils.py
```diff
@@ -466,3 +132,402 @@ def read_geojson(path, **kwargs):
...
+
+    sources: Optional[dict[str, str] | str] = None
+    source_variants: Optional[dict[dict[str, str] | str]] = None
+    variant: str = None
+
+    columns: dict[str, str] = None
+    column_additions: dict[str, str] = EMPTY_DICT
+    column_filters: dict[str, callable] = EMPTY_DICT
+    column_migrations: dict[str, callable] = EMPTY_DICT
+    missing_schemas: dict[str, str] = EMPTY_DICT
```
m-mohr, line 155: Out of curiosity: I've seen this recently in other code, too. Why does python need this? Why not just `var_name = {}`?

fiboa_cli/convert_utils.py
```diff
@@ -466,3 +132,402 @@ def read_geojson(path, **kwargs):
...
+        # It's a bit hacky but allows a gradual conversion from function-based to class-based converters
+        self.__dict__.update({k: v for k, v in kwargs.items() if v is not None})
+        for key in ("id", "short_name", "title", "license", "columns"):
+            assert getattr(self, key) is not None, f"{self.__class__.__name__} misses required attribute {key}"
+
+    @property
+    def ID(self):  # noqa backwards compatibility for function-based converters
+        return self.id
+
+    def migrate(self, gdf):
```
m-mohr, line 171: If I'm not mistaken:

```suggestion
    def migrate(self, gdf) -> gpd.GeoDataFrame:
```

fiboa_cli/convert_utils.py
```diff
@@ -466,3 +132,402 @@ def read_geojson(path, **kwargs):
...
+        return gdf
+
+    def file_migration(self, gdf: gpd.GeoDataFrame, path: str, uri: str, layer: str = None) -> gpd.GeoDataFrame:  # noqa
+        return gdf
+
+    def layer_filter(self, layer, uri):  # noqa
```
m-mohr, line 177: We should probably also add type hints here.

I assume it's the following, but please verify:

```suggestion
    def layer_filter(self, layer: str, uri: str) -> bool:  # noqa
```

fiboa_cli/convert_utils.py
```diff
@@ -466,3 +132,402 @@ def read_geojson(path, **kwargs):
...
+    sources: Optional[dict[str, str] | str] = None
+    source_variants: Optional[dict[dict[str, str] | str]] = None
+    variant: str = None
+
+    columns: dict[str, str] = None
+    column_additions: dict[str, str] = EMPTY_DICT
+    column_filters: dict[str, callable] = EMPTY_DICT
+    column_migrations: dict[str, callable] = EMPTY_DICT
+    missing_schemas: dict[str, str] = EMPTY_DICT
+    extensions: list[str] = tuple()
```
m-mohr, line 156: Shouldn't that better be a set instead of a tuple to avoid duplicates?

fiboa_cli/convert.py
ivorbosloper: Well, there's only a few places. I kept the changes as small as possible so we can do a staged refactoring.

tests/conftest.py
ivorbosloper: No, but it really helped me in getting the class-based approach to work. First write a test, then refactor...

fiboa_cli/convert.py
m-mohr: Yeah, but it would ensure that we remove everything once we actually have all of them converted to class based. Otherwise it could be very fiddly work to figure out what's still needed and what can be removed, I think.

fiboa_cli/__init__.py
ivorbosloper: It's for year. I just thought that **year** might be too specific. E.g. if there's a Converter that handles different regions, or different languages, or something else...

I will change it to an option `--year`...

ivorbosloper: Resolved with https://github.com/fiboa/cli/pull/138/commits/249715ec4dbcae5df844ed9d44dd1eafaf5ccb69

fiboa_cli/convert.py
ivorbosloper: Ah, you want to add comments. Yes, makes sense. Is a `todo` like this good enough? `# todo: remove conditional once class-based converters have been fully implemented`

fiboa_cli/convert.py
ivorbosloper: Yes, will do

fiboa_cli/convert.py
ivorbosloper: See https://github.com/fiboa/cli/pull/138/commits/892c9a4a307ea082c979dd73d7ac771532495b1c

fiboa_cli/datasets/template.py
ivorbosloper: It's not required but it makes it nicer. Will check in one of the tests... https://github.com/fiboa/cli/pull/138/commits/52fd8bce2432066d009e7520e4793d376d22cf1f

fiboa_cli/datasets/template.py
ivorbosloper: Yes. Although this is explicit. 

You can already do this in `BaseConverter.get_urls()` and just forget about the whole sources/source_variants..

fiboa_cli/datasets/template.py
ivorbosloper: Ah, like that. In `SomeConverter.__init__()` you can change the `self.columns`.

fiboa_cli/convert_utils.py
ivorbosloper: Mutability... Now someone can not accidentally change an instance shared by many classes. You will get unexpected bugs...

fiboa_cli/convert_utils.py
ivorbosloper: Will do

fiboa_cli/convert_utils.py
ivorbosloper: Fixed it

fiboa_cli/convert_utils.py
ivorbosloper: I did type `list[str]` but then added instance `tuple()` as a kind-of immutable list. `set()` could also work, than I would initialize it with a `frozenset()`.

tests/conftest.py
ivorbosloper: Pfiew. As our test-suite is still limited I tend to add tests when useful. But I can make a separate pull-requests with most of the current test-additions. I've also added tests for variants, eh years, so they would lead to merge conflicts.

fiboa_cli/datasets/template.py
```diff
@@ -4,170 +4,143 @@
...
+    #     return True
+
+    # Unique identifier for the collection
+    id = "abc"
+    # Geonames for the data (e.g. Country, Region, Source, Year)
+    short_name = "Country, Region, etc."
+    # Title of the collection
+    title = "Field boundaries for Country, Region, etc."
+    # Description of the collection. Can be multiline and include CommonMark.
+    description = """Describe the dataset here."""
```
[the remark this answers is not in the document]
ivorbosloper: Ah, it should? Probably because it ends up in the metadata of the parquet? We can use `f"{self.title} ({self.year})"` when a year is selected...  will add `BaseConverter.get_title()` with this behaviour...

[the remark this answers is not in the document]
ivorbosloper: Yes https://github.com/fiboa/cli/pull/138/commits/6e0c5cb22332df34000ceba62ebc3d44803a063e

fiboa_cli/datasets/template.py
ivorbosloper: Ah, we could keep the original template.py file as `template_convert_function.py` and remove it once everything is refactored... Done.

tests/conftest.py
ivorbosloper: See https://github.com/fiboa/cli/pull/141

ivorbosloper:
So much for a 'minimal' change :)

fiboa_cli/datasets/template.py
m-mohr: That's not what I meant, but is also okay. :-)

fiboa_cli/datasets/template.py
m-mohr: I don't like to do that in init, wouldn't it make more sense to write function for each step which get called and can be overridden?

fiboa_cli/convert.py
m-mohr: Yes, that's excellent.

fiboa_cli/datasets/template.py
ivorbosloper: Sure. You can already use `@property def columns()`. Or we can write a `BaseConverter.get_columns()` that by default returns `self.columns`. I wanted to keep this initial refactor as small as possible. It's nice to be able to use attributes over methods as the code becomes much more concise/readable. Consider the following simplified example:

```
class SomeConverter(BaseConverter):
    id = "dk"
    short_name = "Denmark"
    title = "Denmark Crop Fields (Marker)"
    description = "The Danish Ministry of Food, Agriculture and Fisheries publishes Crop Fields (Marker) for each year."
    extensions = ["https://fiboa.github.io/crop-extension/v0.1.0/schema.yaml"]
    columns = {
        'geometry': 'geometry',
        'Marknr': 'id',
    }
```

Requiring methods for these would end up twice the amount of code and much more ceremony/fuzz:

```
class SomeConverter(BaseConverter):
    def get_id(self):
        return "dk"

    def get_short_name(self): 
        return "Denmark"

    def get_title(self):
        return "Denmark Crop Fields (Marker)"

    def get_description(self):
        return "The Danish Ministry of Food, Agriculture and Fisheries publishes Crop Fields (Marker) for each year."

    def get_extensions(self):
        return ["https://fiboa.github.io/crop-extension/v0.1.0/schema.yaml"]

    def get_columns(sefl):
         return {
            'geometry': 'geometry',
            'Marknr': 'id',
        }
```

I kind of like to only introduce overridable methods if they're useful, so refactor once required.

ivorbosloper: Also, doing year-specific things in a single function (probably in an overridden `Converter.convert()` instead of `__init__`, as `convert` takes the variant) can bundle the business logic instead of spreading this co-related logic out over many diffferent functions. E.g.

```
class SomeConverter(BaseConverter):
    def convert(self, *args, **kwargs):
         if self.year < 2018:
             self.columns["xyz"] = "something"
             self.column_migrations["xyz"] = lambda x: f"{x} {x}"
             self.missing_schemas["properties"]["xyz"] = "string"
         elif self.year < 2022:
             # bundled logic here
         else:
             # bundled logic here

         super().convert(*args, **kwargs)
```


```
class SomeConverter(BaseConverter):
    def get_columns(self):
         if self.year < 2018:
             return old columns
         elif self.year < 2022:
             return intermediate columns
         else:
             return new columns

   def get_missing_schemas(self):
         if self.year < 2018:
            return something with old columns
         elif self.year < 2022:
             return something with intermediate columns
         else:
             return something with new columns

   def get_column_migrations(self):
         if self.year < 2018:
            return something with old columns
         elif self.year < 2022:
             return something with intermediate columns
         else:
             return something with new columns
```

m-mohr: Honestly, for me the spreading out looks much more intuitive to write and understand.

fiboa_cli/convert_utils.py
m-mohr: I think we just have very different approaches in mind. I find it weird that I can't add additional things to the set, which I may want to do depending on inputs.

m-mohr (approved):
Generally LGTM, we just have different opinions on parts of the approach which we can battle out later ;-)

Approving so that we don't run into large merge conflicts and blockers. 

~Please give me some time before merging so that I can do a release before.~ DONE!

fiboa_cli/convert_utils.py
ivorbosloper: Yes, I just fixed the data-sharing issue. Do you want me to make it a set?

fiboa_cli/datasets/template.py
m-mohr: Moved to #146

fiboa_cli/convert_utils.py
m-mohr: Ported to #147

PR #141, opened 2025-01-03
Avoid downloading external data (our tests may fail)

ivorbosloper:


m-mohr (commented):
Looks a bit hacky, but it seems to work.

What's meant with "our tests may fail"?

I'm wondering whether it would be cleaner to use pytest-timeout or so. So we just set a timeout per test of let's say 5 seconds. Usually that should also ensure no external files are downloaded...

ivorbosloper:
Tests should not depend on the content/response-times of websites outside of our control. Using timeouts leads to flaky tests. I wrote this code after an unexpected failing test, for a 500 or 429 error, or a timeout on one of the external websites. I've used the `responses` package previously to mock and limit requests responses, but this code base uses fsspec, not sure how to limit that. And we are accessing urls in the tests, at least the extension schemas..

This code avoids the most common mistakes. We inject the zip files by using a shared 'cache' directory, and if the code changes (e.g. a new year is added) this test will make sure you update the test-content.

m-mohr (approved):

PR #142, opened 2025-01-04
Handle z-dimension automatically

ivorbosloper:
General fix for z-dimension geometries


m-mohr (approved):
Huh, I thought we already had an automatic way of doing this. Seems I was wrong... 👍

PR #143, opened 2025-01-04
LT Euroland

ivorbosloper:
If the Euroland project publishes datasets, we can convert them


fiboa_cli/datasets/commons/euro_land.py
```diff
@@ -0,0 +1,50 @@
+class EuroLandConverterMixin:
+    """
+    Datasets have been published by the
+    [Euroland project](https://europe-land.eu/news/harmonized-database-of-european-land-use-data-published/)
+    as open data. See https://zenodo.org/records/14384070 for a list of open data sets.
+
+    Use this abstract class to create converters based on the euroland repository
+    Subclasses should declare: id, short_name, title, description, providers, license, attribution,
```
m-mohr, line 8: I find the interface as such a bit weak. If it's only clear from a comment what's needed for an implementation, it's hard to understand IMHO. An interface should be something in code logic, a method or abstract class or so...

m-mohr (approved):
The converter is fine, I guess. Some comments for consideration nevertheless.

We're having a different understanding of how to model the interface ;-)

fiboa_cli/datasets/commons/euro_land.py
```diff
@@ -0,0 +1,50 @@
...
+        "geometry": "geometry",
+        "field_id": "id",
+        "farm_id": "farm_id",
+        "crop:code_list": "crop:code_list",
+        "crop_code": "crop:code",
+        "crop_name": "crop:name",
+        'EC_trans_n': 'ec:translated_name',
+        'EC_hcat_n': 'ec:hcat_name',
+        'EC_hcat_c': 'ec:hcat_code',
+        "organic": "organic",
```
m-mohr, line 26: I  haven't had a look at the dataset itself and didn't find a data survey, so I'm wondering what organic may contain? From the field name it sounds like a boolean reather than a string.

fiboa_cli/datasets/commons/euro_land.py
```diff
@@ -0,0 +1,50 @@
...
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        assert self.crop_code_list, f"Please declare a crop_code_list attribute on {self.__class__}"
+        self.column_additions = {**self.column_additions, "crop:code_list": self.crop_code_list}
+        self.providers = [*self.providers, {
+            "name": "Europe-LAND HE Project",
+            "url": "https://doi.org/10.5281/zenodo.14230620",
+            "roles": ["processor"]
+        }]
```
m-mohr, line 50: This way seems rather difficult and unintuitive to me. 

self.providers.append() is what I would expect.

fiboa_cli/datasets/commons/euro_land.py
m-mohr: Fair enough. Should the schema then be uint8 instead of string?

fiboa_cli/datasets/commons/euro_land.py
ivorbosloper: Yes, python is doing this slightly different. Now, I've defined class-members (which are also accessible as instance-members, it's nice but also problematic. Java won't let you do this).

If you want instance members, define them in `__init__()`. I will make a failing test and fix it...

ivorbosloper: Every language has it's quirks. But we can fix this issue in a single place and never worry about it again...

ivorbosloper:
I fixed it in 3 steps;

- write the code how we want it
- show that it won't work without copying the members (with a failing test)
- fix the bugs by copying the shared class-based members in the `BaseConverter.__init__()`. tests work 👍

PR #144, opened 2025-01-07
Make files unix-newlined

ivorbosloper:
These files are windows-based (CR + LF at each line), which shows weird results. Only changed CR+LF to LF with `dos2unix`


m-mohr (approved):
What do you mean with weird results? I thought the days where LF/CR differences don't matter anymore, even Windows Editor can show Linux newlines nowasays. Not the case on Linux? ;-)

ivorbosloper:
> What do you mean with weird results? I thought the days where LF/CR differences don't matter anymore, even Windows Editor can show Linux newlines nowadays. Not the case on Linux? ;-)

Well, it's more consistent like this anyways... I mostly took notice when examing diffs, then all lines would end with `^M`

PR #148, opened 2025-01-08
Manage extensions in converters as a set. fixes #147

### LINKED ISSUE #147: Manage extensions in converters as a set
> Let's make it a set later, because all the non-class-based converters still use lists as well. We can rewrite it in one go....

_Originally posted by @ivorbosloper in https://github.com/fiboa/cli/pull/138#discussion_r1905885738_

ivorbosloper:


m-mohr (approved):
thanks.

PR #149, opened 2025-02-03
Sort geometries by hilbert_distance

ivorbosloper:
Parquet files are great. With an s3-stored file, you can inspect the index and download only the part of your current area of interest. However, this only works efficiently if the geometries in the parquet-blocks are clustered. E.g. if the covering bounding boxes of each parquet-block overlap the entire dataset, you need to download them all for any area of interest. A nice way to achieve clustering is by sorting the geometries by hilbert_distance.

If we use the same function over all published datasets, merging different datasets together could be done efficiently (doing something like a streaming merge-sort merge operation). Let's make it a default operation in the converter.

When sorting on geometry, implicitly the hilbert_distance is used, see https://github.com/geopandas/geopandas/discussions/2893 . This is what I've implemented. However, we might want to customize the parameters for the used hilbert function (e.g. level). The [examples I found](https://notebooksharing.space/view/fb83908941339cc7654dbe271c02b2df49eaf8cc4d434077da58066f9cd1ff86#displayOptions=) didn't work for me, I've tried this:

```
from geopandas.tools.hilbert_curve import _hilbert_distance as hilbert_distance
gdf.sort_values(hilbert_distance(gdf['geometry'], level=15), inplace=True)
gdf.sort_values(gdf['geometry'].hilbert_distance(col, level=15), inplace=True)
gdf.sort_values('geometry', key=lambda col: hilbert_distance(col, level=15), inplace=True)
```

Additionally, do we want to annotate that datasets are sorted this way (in metadata)? Or we could release this in a new version and assume all fiboa-converted datasets after this version are properly sorted...


m-mohr:
👍

> However, we might want to customize the parameters for the used hilbert function (e.g. level).

I looked through the GeoPandas code a bit and the only solution I could find is:

```py
        gdf[":hilbert"] = gdf.hilbert_distance(level=10)
        gdf.sort_values(":hilbert", inplace=True)
        gdf = gdf.drop(columns=[":hilbert"])
```

By the way, I'd place the code in section 4b so that it keeps close to the other code that works on geometries.

> Additionally, do we want to annotate that datasets are sorted this way (in metadata)? Or we could release this in a new version and assume all fiboa-converted datasets after this version are properly sorted...

I don't think we need to add this information to the metadata.

m-mohr:
@ivorbosloper Are you pursuing this further? I just started to make the changes, but then realized that I'm not really sure I understand the level parameter yet, and as such I'm not sure which value would be good here.

ivorbosloper:
> @ivorbosloper Are you pursuing this further? I just started to make the changes, but then realized that I'm not really sure I understand the level parameter yet, and as such I'm not sure which value would be good here. We could also keep the code as it is.

I was just about to ask you the same. Let's keep the code like this for now, the default level is 16. There's a [nice wikipedia explanation on level (order)](https://en.wikipedia.org/wiki/Hilbert_curve), but I have no clue what would be the preferred value for fiboa. This should work for now.

m-mohr (approved):
Okay, I moved the code to step 4b, so this looks good now.

m-mohr:
@ivorbosloper I actually had to move the whole geometry fixing step to after renaming the geometry column. Maybe give it another review and then we can merge.

ivorbosloper:
> @ivorbosloper I actually had to move the whole geometry fixing step to after renaming the geometry column. Maybe give it another review and then we can merge.

Tests work, looks good to me.

PR #152, opened 2025-03-04
Add linting / formatting based on ruff

m-mohr:
Thanks @ivorbosloper - approved from my side.
Please see the one open comment below.

### DIFF
--- a/fiboa_cli/convert_utils.py
+++ b/fiboa_cli/convert_utils.py
@@ -242,12 +242,9 @@ class BaseConverter:
         for uri, target in uris.items():
             is_archive = isinstance(target, list)
             if is_archive:
-                try:
-                    name = name_from_uri(uri)
-                    # if there's no file extension, it's likely a folder, which may not be unique
-                    if "." not in name:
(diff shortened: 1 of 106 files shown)


fiboa_cli/convert_utils.py
```diff
@@ -221,7 +247,7 @@ def download_files(self, uris, cache_folder=None):
                     # if there's no file extension, it's likely a folder, which may not be unique
                     if "." not in name:
                         name = hashlib.sha256(uri.encode()).hexdigest()
-                except:
+                except Exception:  # todo: in which case does this actually happen?
```
m-mohr, line 250: @ivorbosloper Do you happen to know why we do it this way?

name_from_uri would need to throw, but in the function there's already exception handling. So in this case the hashing would need to throw, but in this case it would also throw in the except handler, so that seems to not make sense as it is. What do I miss? Can we remove the exception handling here?

ivorbosloper: Unsure. I pushed the removal commit. All tests work wo it should be fine 🤞 😄

ivorbosloper (approved):

PR #154, opened 2025-03-19
Update lacuna_labels.py

ldemaz:
Changing the id column column to reflect new id variables created in field label geoparquet


ivorbosloper (approved):
Looks good to me (works on new datafile format)

PR #158, opened 2025-03-20
Refactor all function based converters

ivorbosloper:


m-mohr:
What's the status of this, @ivorbosloper? Should we merge this?

ivorbosloper:
@m-mohr  Yes please, we should merge. Then, all function-based converter stuff is gone. This will also be useful when we try to reuse the tooling in vecorel...

fiboa_cli/datasets/ec_ee.py
```diff
@@ -1,83 +1,62 @@
...
-Geospatial Aid Application Estonia Agricultural parcels.
-The original dataset is provided by ARIB and obtained from the INSPIRE theme GSAA (specifically Geospaial Aid Application Estonia Agricultural parcels) through which the data layer Fields and Eco Areas (GSAA) is made available.
-The data comes from ARIB's database of agricultural parcels.
-"""
-PROVIDERS = [
-    {
-        "name": "Põllumajanduse Registrite ja Informatsiooni Amet",
-        "url": "http://data.europa.eu/88u/dataset/pria-pollud",
-        "roles": ["producer", "licensor"],
+class Convert(EuroCropsConverterMixin, BaseConverter):
```
m-mohr, line 10: The old converter sets the year specifically to 2021 in the EC call, I can't find that anymore. Is it still present?

fiboa_cli/datasets/varda.py
```diff
@@ -62,29 +13,57 @@ def MIGRATION(gdf):
...
+        # "effective_until": "datetime:valid_until",
+        # 0000-01-01T00:00:00.000Z and 9999-12-31T00:00:00.000Z should be converted to None
+    }
+
+    # EXTENSIONS = {
+    #    "https://fiboa.github.io/timestamps-extension/v0.1.0/schema.yaml"
+    # }
 
+    def migrate(self, gdf):
+        def convert_from_unit(row, prefix, conversion_factors):
```
m-mohr, line 54: Not a big fan of nested functions, can we move that function a level up?

ivorbosloper: Yes

fiboa_cli/datasets/ec_ee.py
ivorbosloper: There's an updated version of the dataset that does contain the 2021: https://zenodo.org/records/14094196

ivorbosloper:
Would be great to merge... This changelog is annoying :)

fiboa_cli/datasets/ec_ee.py
ivorbosloper: Ah, it get's added to the geoparquet title meta data. I missed this. Just added a fix

PR #161, opened 2025-05-16
Convert Brasil Conab dataset

ivorbosloper:
Implement converter for https://github.com/fiboa/data/issues/46

### DIFF
--- a/fiboa_cli/datasets/br_conab.py
+++ b/fiboa_cli/datasets/br_conab.py
@@ -102,8 +102,10 @@ class Converter(AdminConverterMixin, BaseConverter):
         path, url = key
         path = next(p for p in glob(path) if "_fixed" not in p)
         path2 = path.replace(".shp", "_fixed.shp")
-        datasource_makevalid(path, path2)
-        result[result.index(key)] = (path2, url)
+        if datasource_makevalid(path, path2):
+            result[result.index(key)] = (path2, url)
+        else:
+            result.remove(key)
         return result
 
 
--- a/fiboa_cli/datasets/br_conab.py
+++ b/fiboa_cli/datasets/br_conab.py
@@ -1,5 +1,7 @@
 import math
 
+import numpy as np
+
 from ..convert_utils import BaseConverter
 from .commons.admin import AdminConverterMixin
 
@@ -87,7 +89,7 @@ class Converter(AdminConverterMixin, BaseConverter):
 
     def migrate(self, gdf):
         gdf = gdf.reset_index(drop=True)
-        gdf["area_ha"].combine_first(gdf["Hectares"]).combine_first(gdf["Shape_Area"] / 10000)  #
+        gdf["area_ha"].combine_first(gdf["Hectares"]).replace(np.nan, None, inplace=True)
         gdf.loc[gdf["area_ha"] == 0, "area_ha"] = gdf.to_crs(31982).area / 10000
         gdf["cd_mun"] = gdf["cd_mun"].combine_first(gdf["CD_MUN"]).apply(fformat)
         gdf["nm_mun"] = gdf["nm_mun"].combine_first(gdf["NM_MUN"]).combine_first(gdf["NM_MUNIC"])
--- a/fiboa_cli/datasets/br_conab.py
+++ b/fiboa_cli/datasets/br_conab.py
@@ -1,0 +1,98 @@
+import numpy as np
+import math
+
+from ..convert_utils import BaseConverter
+from .commons.admin import AdminConverterMixin
+
+class Converter(AdminConverterMixin, BaseConverter):
+    _sources = [
+        "Algodao/GO/ALGODAO_GO_Safra_2019_2020.zip",
+        "Algodao/GO/ALGODAO-GO_Safra_2018_2019.zip",
+        "Algodao/GO/GO_ALGODAO_2021.zip",
+        "Algodao/GO/GO_ALGODAO_2223.zip",
+        "Algodao/MS/MS_ALGODAO_2021.zip",
+        "Algodao/MS/MS_ALGODAO_2122.zip",
+        "Arroz_Irrigado/GO/GO_ARROZ_IRRIG_2122.zip",
+        "Arroz_Irrigado/GO/GO_ARROZ_IRRIG_INUND_2324.zip",
+        "Arroz_Irrigado/GO/ARROZ-GO_Safra_2018_2019.zip",
+        "Arroz_Irrigado/MS/ARROZ-MS_Safra_2018_2019.zip",
+        "Arroz_Irrigado/PR/ARROZ-PR_Safra_2017_2018.zip",
+        "Arroz_Irrigado/RS/ARROZ-RS_Safra_2019_2020.zip",
+        "Arroz_Irrigado/SC/ARROZ-SC_Safra_2018_2019.zip",
+        "Arroz_Irrigado/TO/ARROZ-TO_Safra_2017_2018.zip",
+        "Arroz_Irrigado/GO/GO_ARROZ_IRRIG_INUND_2324.zip",
+        "Arroz_Irrigado/TO/TO_ARROZ_IRRIG_2324.zip",
+        "Cana/GO/CANA-GO_Safra_2011_2012.zip",
+        "Cafe/BA/CAFE-BA_Safra_2019.zip",
+        "Cafe/DF/CAFE-DF_Safra_2018.zip",
+        "Cafe/GO/CAFE-GO_Safra_2018.zip",
+        "Cafe/GO/CAFE-GO_Safra_2019.zip",
+        "Cafe/PR/CAFE-PR_Safra_2017.zip",
+        # "Cafe/MG/CAFE-MG_Safra_2017.zip",  # File with invalid geometry
+        "Cafe/DF/DF_CAFE_24.zip",
+        "Cafe/DF/DF_CAFE_24.zip",
+        "Cafe/GO/GO_CAFE_21.zip",
+        "Cafe/RJ/RJ_CAFE_21.zip",
+        "Culturas_de_Verao_1_Safra/DF/CV-DF_Safra_2013_2014.zip",
+        "Culturas_de_Verao_1_Safra/DF/CV-DF_Safra_2014_2015.zip",
+        "Culturas_de_Verao_1_Safra/DF/CV-DF_Safra_2017_2018.zip",
+        "Culturas_de_Verao_1_Safra/TO/CV-TO_Safra_2019_2020.zip"
+    ]
+    sources = {"https://portaldeinformacoes.conab.gov.br/downloads/mapas/"+k: ["*.shp"] for k in _sources}
+    id = "br_conab"
+    short_name = "Conab"
+    title = "Brazil Crop Fields (CONAB)"
+    description = """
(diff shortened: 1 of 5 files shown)


fiboa_cli/datasets/br_conab.py
```diff
@@ -0,0 +1,101 @@
...
+        "Arroz_Irrigado/TO/ARROZ-TO_Safra_2017_2018.zip",
+        "Arroz_Irrigado/GO/GO_ARROZ_IRRIG_INUND_2324.zip",
+        "Arroz_Irrigado/TO/TO_ARROZ_IRRIG_2324.zip",
+        "Cana/GO/CANA-GO_Safra_2011_2012.zip",
+        "Cafe/BA/CAFE-BA_Safra_2019.zip",
+        "Cafe/DF/CAFE-DF_Safra_2018.zip",
+        "Cafe/GO/CAFE-GO_Safra_2018.zip",
+        "Cafe/GO/CAFE-GO_Safra_2019.zip",
+        "Cafe/PR/CAFE-PR_Safra_2017.zip",
+        # "Cafe/MG/CAFE-MG_Safra_2017.zip",  # File with invalid geometry
```
m-mohr, line 31: All of them are invalid or just a couple of them?

fiboa_cli/datasets/br_conab.py
```diff
@@ -0,0 +1,101 @@
...
+        gdf = super().file_migration(gdf, path, uri, layer)
+        # Harmonize projection or pd.concat will fail
+        if gdf.crs.srs != "EPSG:4674":
+            gdf.to_crs(crs="EPSG:4674", inplace=True)
+        return gdf
+
+    def migrate(self, gdf):
+        gdf = gdf.reset_index(drop=True)
+        gdf["area_ha"].combine_first(gdf["Hectares"]).combine_first(gdf["Shape_Area"] / 10000)  #
+        gdf.loc[gdf["area_ha"] == 0, "area_ha"] = gdf.to_crs(31982).area / 10000
```
m-mohr, line 91: Set to null, see JECAM.

fiboa_cli/datasets/br_conab.py
```diff
@@ -0,0 +1,101 @@
...
+    def file_migration(self, gdf, path, uri, layer=None):
+        gdf = super().file_migration(gdf, path, uri, layer)
+        # Harmonize projection or pd.concat will fail
+        if gdf.crs.srs != "EPSG:4674":
+            gdf.to_crs(crs="EPSG:4674", inplace=True)
+        return gdf
+
+    def migrate(self, gdf):
+        gdf = gdf.reset_index(drop=True)
+        gdf["area_ha"].combine_first(gdf["Hectares"]).combine_first(gdf["Shape_Area"] / 10000)  #
```
m-mohr, line 90: Weird dataset :-o

fiboa_cli/datasets/br_conab.py
ivorbosloper: At least one geometry. I didn't find a way to recover from a single error, the script stops at once. This was my resolution for now...

m-mohr: Where does it stop? Can it be made valid with a make_valid call for that specific file in the per-file migration?

fiboa_cli/datasets/br_conab.py
ivorbosloper: Yes, it's a collection of datasets with different attributes. But this harmonizes it....

fiboa_cli/datasets/br_conab.py
ivorbosloper: It stops at `data = gpd.read_file(path, **kwargs)` in convert_utils.py https://github.com/fiboa/cli/blob/main/fiboa_cli/convert_utils.py#L346 . We cannot easily hook into this code. The `geopandas.array.from_wkb(..., on_invalid="raise")` has an option to ignore errors, but we can not change this invocation deeply hidden in the used libraries. There is an [open ticket in geopandas](https://github.com/geopandas/geopandas/issues/2791) to ignore errors.

This is the exception

```
 ...
  File "fiboa/cli/fiboa_cli/__init__.py", line 483, in convert
    convert_(
  File "fiboa/cli/fiboa_cli/convert.py", line 39, in convert
    converter.convert(
  File "fiboa/cli/fiboa_cli/convert_utils.py", line 507, in convert
    gdf = self.read_data(paths, **kwargs)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "fiboa/cli/fiboa_cli/convert_utils.py", line 347, in read_data
    for data, path, uri, layer in self.get_data(paths, **kwargs):
                                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "fiboa/cli/fiboa_cli/convert_utils.py", line 341, in get_data
    data = gpd.read_file(path, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/env/fiboa/lib/python3.12/site-packages/geopandas/io/file.py", line 294, in _read_file
    return _read_file_pyogrio(
           ^^^^^^^^^^^^^^^^^^^
  File "/env/fiboa/lib/python3.12/site-packages/geopandas/io/file.py", line 547, in _read_file_pyogrio
    return pyogrio.read_dataframe(path_or_bytes, bbox=bbox, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/env/fiboa/lib/python3.12/site-packages/pyogrio/geopandas.py", line 318, in read_dataframe
    geometry = from_wkb(geometry, crs=meta["crs"])
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/env/fiboa/lib/python3.12/site-packages/geopandas/array.py", line 215, in from_wkb
    return GeometryArray(shapely.from_wkb(data, on_invalid=on_invalid), crs=crs)
                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/env/fiboa/lib/python3.12/site-packages/shapely/io.py", line 320, in from_wkb
    return lib.from_wkb(geometry, invalid_handler, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
shapely.errors.GEOSException: IllegalArgumentException: Points of LinearRing do not form a closed linestring
```

Alternatively, we could try to fix the source file. Several options come to mind:

### gdal vector make valid

https://gdal.org/en/stable/programs/gdal_vector_geom_make_valid.html#gdal-vector-geom-make-valid , which can be run with `gdal.Run("vector geom make-valid", input=path, output=path2)` from our code. I made an implementation but this does not fix our problem, the geometries remain invalid.

### `ogr2ogr output.shp input.shp -makevalid` 

See [ogr2ogr](https://gdal.org/en/stable/programs/ogr2ogr.html#cmdoption-ogr2ogr-makevalid). However, this is a command line utility and might set some new requirements to your environment. It can be run more directly from python with [pygdaltools](https://pypi.org/project/pygdaltools/), introducing more dependencies. I confirmed works with the CLI.

### ogr/gdal SWIG

Or, skip dependencies and try to work with the osgeo ogr/gdal API. Querying the layer with `select ST_MakeValid(geometry) as geometry, *` with sqlite-dialect from the datasource and writing to a newly created datasource.

ivorbosloper:
Hmmm, the test fails because `from osgeo import ogr` fails. This is the python-gdal binding, installed with `pip install GDAL==<some-version>` . pyogrio, shapely, geopandas, they all use geos/gdal, but we lack a direct python binding is this project.

fiboa_cli/datasets/br_conab.py
m-mohr: I'm trying something right now, but the file download is REALLY slow! ...

ivorbosloper:
@m-mohr I will fix the file, by only if gdal-python bindings are installed. It's awkward to see that 3 different installed libraries use gdal, but using it from python directly seems really hard....

fiboa_cli/datasets/br_conab.py
ivorbosloper: Windows? Or internet challenges? Or maybe they're asleep in Brasil now? 😄

m-mohr:
We don't need it, we can just passt on_invalid = warn. Wait a sec.

fiboa_cli/datasets/br_conab.py
m-mohr: We can use on_invalid via kwargs, it works for me... :-)

m-mohr: I think you misunderstood me, according to the geopandas issue we can pass on_invalid=warn through kwargs in read_file. I tested it just now and it works, see last commit.

Maybe check whether it also works for you and check whether my changes look good, please: https://github.com/fiboa/cli/pull/161/commits/e6b871fcb6bf1a256bc5f46b1c5239d3718c0976

m-mohr:
We don't need the gdal addition, we can just use geopandas :-)
See https://github.com/fiboa/cli/pull/161#discussion_r2104716161

fiboa_cli/datasets/br_conab.py
m-mohr: We could also set it fo fix instead of warn, see https://pyogrio.readthedocs.io/en/latest/api.html#pyogrio.read_dataframe:~:text=on_invalidstr%2C%20optional%20(default%3A%20%E2%80%9Craise%E2%80%9D)

m-mohr:
There's also something wrong with the ids, a lot of them are NaN (and the others are a bit weird with ".0" at the end)

ivorbosloper:
This patch still results in "IllegalArgumentException: Points of LinearRing do not form a closed linestring" on my machine me. I did try this. It might be mac or my specific GDAL version...

```
- Reading layer MG_CAFE_17 into GeoDataFrame
/Users/ivorbosloper/env/fiboa/lib/python3.12/site-packages/pyogrio/raw.py:196: RuntimeWarning: driver ESRI Shapefile does not support open option ON_INVALID
  return ogr_read(
/Users/ivorbosloper/env/fiboa/lib/python3.12/site-packages/pyogrio/raw.py:196: RuntimeWarning: Non closed ring detected. To avoid accepting it, set the OGR_GEOMETRY_ACCEPT_UNCLOSED_RING configuration option to NO
  return ogr_read(
IllegalArgumentException: Points of LinearRing do not form a closed linestring
```

ivorbosloper:
Ah, I got your fix to work

- The [documentation you mentioned](https://pyogrio.readthedocs.io/en/latest/api.html#pyogrio.read_dataframe:~:text=on_invalidstr%2C%20optional%20(default%3A%20%E2%80%9Craise%E2%80%9D)) mentions the `on_invalid` parameter requires `shapely>=2.1`
- The pyogrio engine needs to be higher than 0.8.0 (does not work) or you should use fiona.

I will upgrade the shapely dependency. Should we introduce a `pyogrio>=0.11.0` dependency of is pyogrio optional?

ivorbosloper:
> Yeah, we should add the shapely dependency if we need require more then our upstream dependency.
> 
> Not sure about pyiogrio. We use it as default, but if we dont require it to be a specific version, we can just let upstream handle it.
> 
> I think we need to require Python 3.10 from now on, which is fine.

I moved up to testing python 3.11, and it works...

PR #164, opened 2025-05-19
Add Jecam converter

ivorbosloper:
Implements https://github.com/fiboa/data/issues/47


fiboa_cli/datasets/jecam.py
```diff
@@ -0,0 +1,66 @@
...
+            "crop:name": {"type": "string"},
+            "crop:code_list": {"type": "string"},
+            "irrigated": {"type": "boolean"},
+        }
+    }
+
+    def migrate(self, gdf) -> gpd.GeoDataFrame:
+        gdf = super().migrate(gdf)
+        # Pick some CRS for missing area's, 32631 (central afrika) or 3857
+        gdf.loc[gdf["Area_ha"] == 0, "Area_ha"] = gdf.to_crs(32631).area / 10000
```
m-mohr, line 64: Didn't we centralize this? We have the improve command and I think we did decide to not implement this in the converters individually, right?

ivorbosloper: Yes, agreed. However, the dataset is invalid (`fiboa validate output.parquet --data`) because the fiboa schema forbids `area==0`. So I can leave the area column out (throwing away the area-data that's correct in most rows) or fill the 0-areas with a geometry-area calculation....

ivorbosloper: I'm not sure if `Null/None/NaN` is allowed...

m-mohr: null (Python None) is allowed.

fiboa_cli/datasets/jecam.py
```diff
@@ -0,0 +1,65 @@
...
+        }
+    ]
+    attribution = "JECAM SIRAD, https://doi.org/10.18167/DVN1/P7OLAP"
+    license = "CC-BY-4.0"
+    columns = {
+        "geometry": "geometry",
+        "Id": "id",
+        "Area_ha": "area",
+        "AcquiDate": "determination_datetime",
+        "Country": "country_name",
```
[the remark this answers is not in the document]
ivorbosloper: Yes, but is requires a mapping I would have to make...

m-mohr (approved):
Thanks.

PR #165, opened 2025-05-20
Simplify command error logging

ivorbosloper:
Simplify code and debugging


m-mohr:
I've got a different implementation in vecorel CLI that solves the same issue. Do we still want to change it here or wait for the upstream change?

ivorbosloper:
Ah, let's focus on Vecorel. I just need an Exception+Trace when implementing/debugging new converters...

PR #171, opened 2025-07-29
India 10k

ivorbosloper:
Add india_10k dataset converter


m-mohr (approved):

ivorbosloper:
> Should we add the admin extension?

Yes, done. Additionally, I noticed the `area` was defined in m² instead of ha and I (finally) set up a generic/simple method for fixing that...

m-mohr:
So should we apply this to all other converters, too? :-D

Issue #172, opened 2025-07-31
Simplify scaling area attribute (e.g. from square meter to hectare)


ivorbosloper:
We created a [simplified/generic method for scaling provided area attribute values]  https://github.com/fiboa/cli/pull/171/commits/b737e6b12b2d0f5dde034924f4fdd773a95384bb . We should apply this for all converters.

_Originally posted by @m-mohr in https://github.com/fiboa/cli/issues/171#issuecomment-3137846755_

PR #173, opened 2025-08-01
Simplify scaling area attribute #172

ivorbosloper:
Implements #172

### DIFF
--- a/fiboa_cli/convert_utils.py
+++ b/fiboa_cli/convert_utils.py
@@ -483,6 +483,7 @@ class BaseConverter:
                     gdf[area_key] == 0, gdf.area * self.FACTOR_M2_TO_HA, gdf[area_key]
                 )
             else:
+                assert gdf.crs.to_dict()['units'] == 'm', f"Dataset projection is not meter-based: {gdf.crs}"
                 gdf[area_key] = gdf.area * self.FACTOR_M2_TO_HA
 
         gdf = self.post_migrate(gdf)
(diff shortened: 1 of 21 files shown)


fiboa_cli/convert_utils.py
```diff
@@ -472,10 +473,18 @@ def convert(
...
             gdf[area_key] *= self.area_factor
 
+        if self.area_fill_zero:
+            if area_key in gdf.columns:
+                gdf[area_key] = np.where(
+                    gdf[area_key] == 0, gdf.area * self.FACTOR_M2_TO_HA, gdf[area_key]
+                )
+            else:
+                gdf[area_key] = gdf.area * self.FACTOR_M2_TO_HA
+
```
m-mohr, line 487: gdf,area may be invalid for a geographic CRS using degrees as units; we'd need to use [GeoSeries.to_crs()](https://geopandas.org/en/stable/docs/reference/api/geopandas.GeoSeries.to_crs.html#geopandas.GeoSeries.to_crs) to project geometries to a planar CRS before using this function. I think we do this already somewhere else in the code, potentially in the improve command, maybe something to centralize.

m-mohr: That's fine, but maybe we should error in case the flag is set and doesn't make sense...

ivorbosloper: Hmmm, I would have to check if we can get the SRID-definition. It's in the `proj` library for certain. Will check it out.

m-mohr: See https://github.com/fiboa/cli/blob/main/fiboa_cli/improve.py#L85

ivorbosloper: Easier than expected: https://github.com/fiboa/cli/pull/173/commits/06f0d895ba476b20ff636c4e0d9f6cd8263f07c9

PR #176, opened 2025-08-27
Migrate to Vecorel and fiboa 0.3

m-mohr:
When do we want to merge?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+## [v0.20.2] - 2025-08-29
+
 - Reimplementation tests and CI
 - Update extensions namespace to fiboa.org, use vecorel schemas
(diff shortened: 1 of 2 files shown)


ivorbosloper:
We need vecorel@0.2.4 (support for int32 in json-schemas)

m-mohr:
v0.2.4 has been released

ivorbosloper:
Tests fail on other os. Should we fix it or skip "windows-latest" and "macos-latest" in https://github.com/fiboa/cli/pull/176/commits/aaec72dbf91dece9525b380e539bbb21c6888061

m-mohr:
Interesting, works on Linux, fails on Windows and Mac...

m-mohr:
So Windows is an issue that I also got on Windows in Vecorel CLI, you need to create the temporary files a little differently. I can investigate that.

MacOS, I can't investigate. It seems Pixi setup isn't working. That seems to be something GitHub specific?

ivorbosloper (approved):
Yes

m-mohr:
Fixed the MacOS build and removed some Python versions to limit the resource usage.

PR #198, opened 2025-12-08
Implement DuckDB-based converter

ivorbosloper:
The downside of our geopandas-based FiboaConverters is that the dataframe need to be in memory. This can be huge. A completely different approach would be to use duckdb for conversion. To test it's viability, I took the dataset (Japan) with largest dataframe. I don't need a 128GB+ machine anymore, it's fast & performant on a laptop using 20GB memory and runs way faster (<30minutes).

I need some help with the parquet metadata...


ivorbosloper:
I think duckdb parquet writer doesn't support setting the nullability derived from the resultset.

```
$ duckdb
D CREATE TABLE test(id varchar not null);
D insert into test values ('A'), ('B');
D COPY (select * from test) to '/tmp/test.parquet' (FORMAT PARQUET);
D exit

$ parquet-dump-schema /tmp/test.parquet 
required group field_id=-1 duckdb_schema {
  optional binary field_id=-1 id (String);
}
```

same for `COPY test to '/tmp/test.parquet' (FORMAT PARQUET)`

m-mohr:
For me that sounds like a bug in duckdb. Is there an open issue for it, otherwise maybe open one?

m-mohr:
@ivorbosloper In which state is this PR? Was the collection-level metadata and column schema ever solved? Is this generating valid fiboa/vecorel files?

m-mohr:
Updated PR to work with new duckdb implementation in vecorel-cli, needs a release of vecorel-cli first.

ivorbosloper:
Superseded. The converter now lands in two places: `FiboaDuckDBBaseConverter` in #272 (rebased on #266, with the JP converter that uses it in #273), and the generic version m-mohr moved upstream in vecorel/cli#29. Once vecorel/cli#29 is released, fiboa's copy goes away and #273 imports the upstream class. Closing this one so the three of them do not compete.

PR #207, opened 2026-05-19
PerFileBaseConverter

ivorbosloper:
Per-file migration and merging the result, decreasing memory requirements for large data sets.

Depends on a new release of Vecorel


ivorbosloper:
@m-mohr There's a highly relevant part here that fixes the hilbert-sort . Would be great to see this on main before other people start publishing new data sets in the wrong sort-order...

m-mohr:
> @m-mohr There's a highly relevant part here that fixes the hilbert-sort . Would be great to see this on main before other people start publishing new data sets in the wrong sort-order...

Isn't that primarily in vecorel CLI? I'll try to review and merge those first, then create a release...

PR #219, opened 2026-08-29
Update vecorel-cli to v0.2.16, require Python env accordingly, etc.

m-mohr:
Should fix the hilbert issue mentioned in #207


ivorbosloper (approved):
Yes, would be great to see this merge

